PR #173, opened 2022-09-28
PEAR- 570 Facet Factory

craigrbarnes:
## Description

- Adds a facet component factory that will use the `facet_type` to correctly create a facet component
- Facet now requires a hooks prop that defines the get/set function needed by that facet. This supports using the Facet components for both Core and app-based filtering.
Also implements:
- PEAR-406
- PEAR-552
- PEAR-554

## Checklist

- [x] Added proper unit tests

image
image


paribartandhakal (approved):
Code wise it looks good to me.

PR #331, opened 2022-12-15
PEAR-856: do not render children if closed or not active

craigrbarnes:
## Description
Changes  CollaspableContainer to no render children when collapsed
Unloads CohortManager tabs (Summary View| Cases View)

Both prevent component updates and data fetched when not visible.
## Checklist

- [ ] Added proper unit tests


paribartandhakal (approved):
Feels very snappy. I double-checked the network tabs too. Simple fix for great results! 👍 
Should we be doing this in other places too if applicable?

PR #561, opened 2023-05-31
PEAR-918: fix survival plot being blank in mutation frequency

alilahrime:
## Description
[pear-918](https://jira.opensciencedatacloud.org/browse/PEAR-918)
## Checklist

- [ ] Added proper unit tests
      flagged issues


craigrbarnes (commented):
See my comment on the data length test. Note that it is breaking a CohortComparision test which is why the pipeline fails. I pushed a change to confirm this is the cause as

PR #691, opened 2023-08-28
PEAR 1322 / 1621 - Mutations button, survival plot in Mutations tab not plotting for a mutation

alilahrime:
## Description
[pear-1322](https://gdc-ctds.atlassian.net/browse/PEAR-1322)


alilahrime:
> Oops - I didn't see that this had more commits.
> 
> This isn't working as expected.
> 
> * use a cohort with bronchus and lung primary site
> * in Mutation Frequency, Genes tab, click on the # Mutations button for TP53
> * it switches to the Mutations tab, applies the search, but it's not attempting to plot for the topmost mutation
> 
> Screenshot 2023-10-23 at 12 21 15 PM

it should plot topmost now afterwards; was missing a filter

https://github.com/NCI-GDC/gdc-frontend-framework/assets/90918464/852dd142-7a36-48cc-b781-55ec0a8ad154

wteouchicago:
working as expected. 👍

packages/portal-proto/src/features/genomic/GenesAndMutationFrequencyAnalysisTool.tsx
```diff
@@ -190,7 +192,8 @@ const GenesAndMutationFrequencyAnalysisTool: React.FC = () => {
       setAppMode("ssms");
       setSearchTermsForGeneId({ geneId: geneId, geneSymbol: geneSymbol });
     },
-    [],
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [cohortFilters, genomicFilters],
```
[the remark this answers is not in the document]
alilahrime: you're right. this was left in by mistake, good catch

packages/portal-proto/src/features/genomic/GenesAndMutationFrequencyAnalysisTool.tsx
```diff
@@ -69,6 +69,8 @@ const GenesAndMutationFrequencyAnalysisTool: React.FC = () => {
   const prevTopGeneSSMS = usePrevious(topGeneSSMS);
   const prevAppMode = usePrevious(appMode);
 
+  // random change
+
```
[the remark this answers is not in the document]
alilahrime: removed

packages/portal-proto/src/features/GenomicTables/GenesTable/utils.tsx
```diff
@@ -309,6 +309,7 @@ export const useGenerateGenesTableColumns = ({
         cell: ({ row }) => {
           const count = row.original["#_mutations"] ?? 0;
           const disabled = count === 0;
+          const { symbol = "", label: name = "" } = row?.original["survival"];
```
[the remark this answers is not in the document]
alilahrime: seems like this was removed from merging develop. ty for pointing out

craigrbarnes:
I noticed one oddity that @wteouchicago might want to confirm is an issue.

MF App: All GDC Cohort, default filtering

1. Select a Mutation in the table: `PTEN` 
2. Tab switches to Mutations shows the top Mutation as `PTEN R130Q`; search is filled in with `ENSG00000171862`
3. Clear search
4. The top mutation table changes to `BRAF 640B`; however, the survival plot still shows `PTEN R130Q`

wteouchicago:
> I noticed one oddity that @wteouchicago might want to confirm is an issue.
> 
> MF App: All GDC Cohort, default filtering
> 
> 1. Select a Mutation in the table: `PTEN`
> 2. Tab switches to Mutations shows the top Mutation as `PTEN R130Q`; search is filled in with `ENSG00000171862`
> 3. Clear search
> 4. The top mutation table changes to `BRAF 640B`; however, the survival plot still shows `PTEN R130Q`

It's a bit weird -- ideally it would plot the top mutation, but I think we can live with it unless it's a super easy fix. I can justify it if necessary and/or we can fix it later. 
I found some weirdness with top mutation plotting recently: https://gdc-ctds.atlassian.net/browse/PEAR-1621 -- maybe they are related and can be fixed with PEAR-1621? 🤷‍♀️

alilahrime:
> > I noticed one oddity that @wteouchicago might want to confirm is an issue.
> > MF App: All GDC Cohort, default filtering
> > 
> > 1. Select a Mutation in the table: `PTEN`
> > 2. Tab switches to Mutations shows the top Mutation as `PTEN R130Q`; search is filled in with `ENSG00000171862`
> > 3. Clear search
> > 4. The top mutation table changes to `BRAF 640B`; however, the survival plot still shows `PTEN R130Q`
> 
> It's a bit weird -- ideally it would plot the top mutation, but I think we can live with it unless it's a super easy fix. I can justify it if necessary and/or we can fix it later. I found some weirdness with top mutation plotting recently: https://gdc-ctds.atlassian.net/browse/PEAR-1621 -- maybe they are related and can be fixed with PEAR-1621? 🤷‍♀️

tried to resolve, but attempting to implement solution to plot the topmost after clearing the search led to unintended side effects (plotting/unplotting whenever the search term changes and then no longer retaining the default behavior of plotting topmost when button is unselected)

wteouchicago:
> tried to resolve, but attempting to implement solution to plot the topmost after clearing the search led to unintended side effects (plotting/unplotting whenever the search term changes and then no longer retaining the default behavior of plotting topmost when button is unselected)

that's ok -- we can attempt to fix later

wteouchicago:
still looks good for 1322 and also checked that PEAR-1621 is fixed by this. 👍

paribartandhakal:
I see 2 linter warnings regarding dependency array. Can we fix those? Suppressing might not be an ideal solution.

alilahrime:
> I see 2 linter warnings regarding dependency array. Can we fix those? Suppressing might not be an ideal solution.

thank you, i agree. its fixed now. updated dependency arrays for 2 warning sources & still functions as before

packages/portal-proto/src/features/GenomicTables/SomaticMutationsTable/SMTableContainer.tsx
```diff
@@ -164,8 +167,46 @@ export const SMTableContainer: React.FC<SMTableContainerProps> = ({
...
+    topSSM,
+    previousTopSSM,
+    handleSurvivalPlotToggled,
+    searchTermsForGene?.geneSymbol,
+  ]);
+
+  useEffect(() => {
+    if (
+      topSSM &&
+      !isEqual(selectedSurvivalPlot, previousSelectedSurvivalPlot) &&
```
paribartandhakal, line 183: Just a suggestion: If you use `useDeepCompareEffect` from `use-deep-compare`, you don't have to check these manually. We have been using this library for these scenarios.

alilahrime: alright, using useDeepCompare now

packages/portal-proto/src/features/GenomicTables/SomaticMutationsTable/SMTableContainer.tsx
```diff
@@ -164,8 +164,42 @@ export const SMTableContainer: React.FC<SMTableContainerProps> = ({
         "gene.ssm.ssm_id",
       );
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [topSSM]);
+  }, [
+    data,
```
craigrbarnes, line 168: eslint warning:` useDeepCompareCallback has unnecessary dependencies: 'data' and 'searchTerm'. Either exclude them or remove the dependency array`

So remove data and searchTerm

craigrbarnes (dismissed):

packages/portal-proto/src/features/GenomicTables/SomaticMutationsTable/SMTableContainer.tsx
craigrbarnes: Pushed the change

craigrbarnes (commented):
Fixed requested change

packages/portal-proto/src/features/GenomicTables/SomaticMutationsTable/SMTableContainer.tsx
```diff
@@ -145,13 +145,13 @@ export const SMTableContainer: React.FC<SMTableContainerProps> = ({
         geneSymbol: geneSymbol,
         genomicFilters: genomicFilters,
         cohortFilters: cohortFilters,
-        caseFilter: { mode: "", root: {} } as FilterSet,
+        caseFilter: caseFilter,
       });
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [searchTermsForGene, genomicFilters, cohortFilters]);
+  }, [searchTermsForGene, genomicFilters, cohortFilters, caseFilter]);
```
paribartandhakal, line 152: I am not sure I understand how `getTopSSM` works in terms of reference equality. Does `useGetSsmTableDataMutation` return the same one every render?

cc. @amlehman @craigrbarnes @alilahrime

alilahrime: I agree with Craig. useMutation [docs](https://redux-toolkit.js.org/rtk-query/api/created-api/hooks#usemutation) says the hook subscribes the component to the result and re-renders only as status/data changes. Even if there was multiple getTopSSM references, they should still subscribe to the same result and shouldn't cause unnecessary re-renders by being included in dependency array

packages/portal-proto/src/features/GenomicTables/SomaticMutationsTable/SMTableContainer.tsx
```diff
@@ -164,8 +164,36 @@ export const SMTableContainer: React.FC<SMTableContainerProps> = ({
         "gene.ssm.ssm_id",
       );
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [topSSM]);
+  }, [topSSM, handleSurvivalPlotToggled, searchTermsForGene?.geneSymbol]);
+
+  useDeepCompareEffect(() => {
```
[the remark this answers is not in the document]
alilahrime: I agree that less redundant code is nice, however that caused an infinite loop. i reverted changes to retain functionality and included some changes you suggested for readability.

packages/portal-proto/src/features/GenomicTables/SomaticMutationsTable/SMTableContainer.tsx
```diff
@@ -145,27 +148,55 @@ export const SMTableContainer: React.FC<SMTableContainerProps> = ({
...
             })}`
-          : "",
-        "gene.ssm.ssm_id",
-      );
-    }
+          : "";
+        handleSurvivalPlotToggled(ssm_id, description, "gene.ssm.ssm_id");
+      }
+    },
     // eslint-disable-next-line react-hooks/exhaustive-deps
```
paribartandhakal, line 189: here too

paribartandhakal: Was it already re-rendering indefinitely before my suggested changes?

packages/portal-proto/src/features/GenomicTables/SomaticMutationsTable/SMTableContainer.tsx
```diff
@@ -145,27 +148,55 @@ export const SMTableContainer: React.FC<SMTableContainerProps> = ({
...
-        ssm_id,
-        consequence_type
+      const description = consequence_type
+        ? `${searchTermsForGene?.geneSymbol ?? ""} ${aa_change} ${humanify({
+            term: consequence_type.replace("_variant", "").replace("_", " "),
+          })}`
+        : "";
+      handleSurvivalPlotToggled(ssm_id, description, "gene.ssm.ssm_id");
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
alilahrime: yeah i meant to add that back

packages/portal-proto/src/features/GenomicTables/SomaticMutationsTable/SMTableContainer.tsx
alilahrime: i didn't notice it after adding useDeepCompare, but it might've been because of handleSurvivalPlotToggled being included in deps array.

paribartandhakal: Yes, that was to consolidate the condition of other useEffect you have on Line 157.

So what I was trying to was combine two useEffect's logic in one as they had the same bit of code. They are both calling `handleSurvivalPlotToggled` based on two different conditionals, right?

alilahrime: 1 effect has the selected survival plot as a dependency while the other does not. the combined effect updates the selected plot either way so thats why its looping i think

craigrbarnes (approved):
How the suvivial plot is set really needs refactoring (PEAR-1657). As this is working and needed for 2.11.0 I will approve. But we will need to revist this.

PR #742, opened 2023-09-27
PEAR-1466: styling theming adjustments

cemile-ctds:
## Description
[PEAR-1466](https://gdc-ctds.atlassian.net/browse/PEAR-1466) branch includes various styling and minor adjustments to the Analysis Center, Projects, Cohort Builder, and Repository to ensure a consistent and polished UI.

This branch also includes the task for UX/UI improvements in the Cohort Builder [PEAR-1358](https://gdc-ctds.atlassian.net/browse/PEAR-1358).

### DIFF
--- a/packages/portal-proto/src/features/cohortBuilder/FacetTabs.tsx
+++ b/packages/portal-proto/src/features/cohortBuilder/FacetTabs.tsx
@@ -316,7 +316,7 @@ export const FacetTabs = (): JSX.Element => {
         onTabChange={setActiveTab}
         keepMounted={false}
         classNames={{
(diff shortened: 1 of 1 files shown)


ocshawn:
we need to revisit this as the use of an ID "liveRegion" in a component will lead to multiple ids with the same name on the page, code not working as expected.

ocshawn:
also by having text in a liveRegion on page load that text is read when user enters the page

craigrbarnes:
> we need to revisit this as the use of an ID "liveRegion" in a component will lead to multiple ids with the same name on the page, code not working as expected.

Since we should fix the id issue sooner I will open up 2 tickets to address these issues.

PR #778, opened 2023-10-25
PEAR-1441: add tooltip to table search with unique styling

ocshawn:
## Description
- add tooltip to table search with unique styling

## Checklist

- [x] Added proper unit tests
      flagged issues

Screenshot from 2023-10-26 17-30-55


ocshawn:
ready to look at again, please let me know if you find more tables or modals where its not working correctly.

wteouchicago:
I'm seeing the "tooltip" showing up as expected now.

Spotted one thing that we can probably fix later if it's a pain -- see screenshot. The tooltip is appearing over the header of summary pages when the pages are loaded in modals.

Screenshot 2023-11-03 at 12 17 34 AM

ocshawn:
yes it is a known issue, another option would be to go with a fix that does not have any accessibility, as the frameworks need to be updated.

cemile-ctds:
The tooltip text size appears to be set as text-sm, but I couldn't confirm whether something is overwriting this class because it seems larger than the placeholder text.

Additionally, the font-family is currently Montserrat. We should use Noto Sans for tooltips.

ocshawn:
font and font size fixed, good catch

packages/portal-proto/src/components/Table/VerticalTable.tsx
```diff
@@ -169,6 +177,25 @@ function VerticalTable<TData>({
     handleChange({ newSearch: "" });
   };
 
+  const tooltipContainer = search?.tooltip
```
paribartandhakal, line 180: Nitpick: Component name should be PascalCased. 
```suggestion
  const TooltipContainer = search?.tooltip
```

packages/portal-proto/src/features/GeneSummary/GeneSummary.tsx
```diff
@@ -280,6 +280,7 @@ const GeneView = ({
                   cohortFilters={cohortFilters}
                   genomicFilters={genomicFilters}
                   isModal={isModal}
+                  inModal={isModal}
```
paribartandhakal, line 283: I feel like `isModal` is sufficient and we don't need a new prop`inModal`. Is there any reason why you added this when there is already `isModal`?

ocshawn: from what i see `isModal` is for opining links in modals where `inModal` is for identifying if you are currently in a modal

ocshawn: this is not the only place that the component is called, in other locations it has separate values

PR #795, opened 2023-11-10
fix(PEAR-1615): resolve summary page crashing

alilahrime:
[pear-1615](https://gdc-ctds.atlassian.net/browse/PEAR-1615)

### DIFF
--- a/packages/portal-proto/src/features/GenomicTables/SomaticMutationsTable/SMTableContainer.tsx
+++ b/packages/portal-proto/src/features/GenomicTables/SomaticMutationsTable/SMTableContainer.tsx
@@ -142,15 +142,19 @@ export const SMTableContainer: React.FC<SMTableContainerProps> = ({
   const [getTopSSM, { data: topSSM }] = useGetSsmTableDataMutation();
 
   useEffect(() => {
-    getTopSSM({
-      pageSize: 1,
-      offset: 0,
-      searchTerm: searchTermsForGene?.geneId ?? "",
-      geneSymbol: searchTermsForGene?.geneSymbol ?? "",
-      genomicFilters: genomicFilters,
-      cohortFilters: cohortFilters,
-      caseFilter: { mode: "", root: {} } as FilterSet,
-    });
+    if (searchTermsForGene) {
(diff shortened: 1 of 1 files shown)


craigrbarnes (approved):
Change look fine. No longer seing the error and the search works as expected

wteouchicago:
No longer crashing on gene page or case page.

Some issues:

1. I noticed in the mutations table on the gene and case page that the checkboxes, buttons, and subrow expansion no longer work. I can't tell if it's related to this PR or not since it's crashing in develop for me. Let me know if it's not related and I'll file a separate bug. 
2. I noticed some slowness so I looked in the network tab. There appears to be infinite(?) graphql calls for the ssmTable. This may be the cause of 1. I didn't see this happening in prod-uat but can't test develop.

alilahrime:
> No longer crashing on gene page or case page.
> 
> Some issues:
> 
> 1. I noticed in the mutations table on the gene and case page that the checkboxes, buttons, and subrow expansion no longer work. I can't tell if it's related to this PR or not since it's crashing in develop for me. Let me know if it's not related and I'll file a separate bug.
> 2. I noticed some slowness so I looked in the network tab. There appears to be infinite(?) graphql calls for the ssmTable. This may be the cause of 1. I didn't see this happening in prod-uat but can't test develop.


> No longer crashing on gene page or case page.
> 
> Some issues:
> 
> 1. I noticed in the mutations table on the gene and case page that the checkboxes, buttons, and subrow expansion no longer work. I can't tell if it's related to this PR or not since it's crashing in develop for me. Let me know if it's not related and I'll file a separate bug.
> 2. I noticed some slowness so I looked in the network tab. There appears to be infinite(?) graphql calls for the ssmTable. This may be the cause of 1. I didn't see this happening in prod-uat but can't test develop.

1. I can confirm that the infinite graphql calls were related to the changes, and i think that it caused the other issues if the subrow buttons were clicked first. i don't experience issues with other buttons/checkboxes anymore too after clicking the subrow expansion. 

2. makes sense i think the slowness is related to 1) and should be resolved now.

wteouchicago:
looks good 👍

PR #802, opened 2023-11-15
PEAR-1544: Table checkbox accessibility

alilahrime:
[pear-1544](https://gdc-ctds.atlassian.net/browse/PEAR-1544)


cemile-ctds:
1- The “checkbox for selecting projects table row GENIE-MSK” is a bit lengthy. Could we simplify it to "Select the GENIE-MSK project row"?
ProjectsTable

2- There seems to be a bug here; the variable is not being replaced with the name.
FaceCards

3- We don't use the correct information for the tables listed below.
```
cDAVE: well, the name (but it's under the header <property name>)
Mutation Frequency > genes table: Symbol
Mutation Frequency > mutations table: Protein Change
```

4- The set operation table announces 'checkbox for selecting table row' when the screen reader focuses on the checkbox. Please replace it with 'Select the {{Set Operations Column value}} row.'

packages/portal-proto/src/features/GenomicTables/GenesTable/utils.tsx
```diff
@@ -56,11 +56,12 @@ export const useGenerateGenesTableColumns = ({
             classNames={{
               input: "checked:bg-accent checked:border-accent",
             }}
+            aria-label={`checkbox for selecting genes table rows currently displayed`}
```
ocshawn, line 59: this is unhelpful to someone who cannot see the page

packages/portal-proto/src/features/GenomicTables/GenesTable/utils.tsx
```diff
@@ -56,11 +56,12 @@ export const useGenerateGenesTableColumns = ({
...
+            aria-checked={table.getIsAllRowsSelected()}
```
ocshawn, line 60: this is not needed its handled automatically by the Checkbox

packages/portal-proto/src/features/GenomicTables/GenesTable/utils.tsx
```diff
@@ -69,7 +70,8 @@ export const useGenerateGenesTableColumns = ({
             classNames={{
               input: "checked:bg-accent checked:border-accent",
             }}
-            aria-label="checkbox for selecting table row"
+            aria-label={`checkbox for selecting genes table row ${row.id}`}
```
ocshawn, line 73: consider using aria-labelledby instead, row.id is also not helpful as a label consider using something like symbol. Also your wording reads as 'checkbox for selecting genes table row ${row.id} checkbox not checked' this should be simplified. 'checkbox for selecting genes table row' will be read on every row and does not tell the user anything new

ocshawn (commented):
my three comments can be applied to all of the items

packages/portal-proto/src/features/GenomicTables/GenesTable/utils.tsx
alilahrime: should we just conditionally change the aria-label based off checked status?

packages/portal-proto/src/features/GenomicTables/GenesTable/utils.tsx
alilahrime: true, changed to include comma separated list of the unique identifiers

packages/portal-proto/src/features/GenomicTables/GenesTable/utils.tsx
alilahrime: using symbol now within the aria-label & shortened wording to be less repetitive.

ocshawn (commented):
i'm concerned with the use of static IDs, will these components ever be on the page more than once?

craigrbarnes:
> i'm concerned with the use of static IDs, will these components ever be on the page more than once?

I agree: EnumFacet, for example, will have multiple copies on one page.

packages/portal-proto/src/features/GenomicTables/GenesTable/utils.tsx
ocshawn: ideally you would want something like "select all rows on page x of x"

ocshawn (dismissed):
ignore this its space bar for checked boxes :facepalm: 

below to document my dumb comment 
_cannot select anything in projects table with keyboard, tried using enter key it did nothing_

alilahrime:
> > i'm concerned with the use of static IDs, will these components ever be on the page more than once?
> 
> I agree: EnumFacet, for example, will have multiple copies on one page.

I see your point, the EnumFacet is a good generic example. So instead of static id would `aria-labelledby={`selectFacet-${value}`}` suffice for example?

ocshawn:
@alilahrime  For Ids consider using the react useId https://react.dev/reference/react/useId append that value to your existing ids to make them unique to that components instance

craigrbarnes:
> @alilahrime For Ids consider using the react useId https://react.dev/reference/react/useId append that value to your existing ids to make them unique to that components instance

That is a good use of useId()

packages/portal-proto/src/features/set-operations/SetOperationTable.tsx
```diff
@@ -154,6 +161,7 @@ export const SetOperationTable = ({
 
   return (
     <VerticalTable
+      aria-hidden="true"
```
craigrbarnes: Why is this being hidden?

craigrbarnes (changes requested):
There are issues with the use of static ids: 
  For example ,in Set Operations table, if I select S1 intersect S2 and then go to either of the other 2 checkboxs, the screen reader will state "uncheck S1 intersect S2" which is incorrect.

craigrbarnes:
> > @alilahrime For Ids consider using the react useId https://react.dev/reference/react/useId append that value to your existing ids to make them unique to that components instance
> 
> That is a good use of useId()

No problem, I respect your knowledge on this.

packages/portal-proto/src/features/set-operations/SetOperationTable.tsx
alilahrime: removed

alilahrime:
> There are issues with the use of static ids: For example ,in Set Operations table, if I select S1 intersect S2 and then go to either of the other 2 checkboxs, the screen reader will state "uncheck S1 intersect S2" which is incorrect.

the checked/unchecked part is being handled by the mantine Checkbox component and i wasnt able to override

packages/portal-proto/src/components/Modals/SetModals/SavedSets.tsx
```diff
@@ -78,17 +78,26 @@ const SavedSets: React.FC<SavedSetsProps> = ({
...
+              classNames={{
+                input: "checked:bg-accent checked:border-accent",
+              }}
+              aria-labelledby={"savedSetsSelectAll"}
+              {...{
+                checked: row.getIsSelected(),
+                onChange: row.getToggleSelectedHandler(),
+              }}
+            />
+          </>
```
ocshawn: don't need to create a separate hidden label you should place an id on the name column and use aria-labelledby to point to that

ocshawn: also dont need to say `Select` : `Unselect` in the text the screen reader will handle that based on the state of the checkbox

packages/portal-proto/src/components/Modals/SetModals/SetTable.tsx
ocshawn, line 1: what was wrong with the old way, did it work? seems to be what the jira is asking for

packages/portal-proto/src/features/GenomicTables/GenesTable/utils.tsx
```diff
@@ -50,30 +50,49 @@ export const useGenerateGenesTableColumns = ({
...
-          <Checkbox
-            size="xs"
-            classNames={{
-              input: "checked:bg-accent checked:border-accent",
-            }}
-            {...{
-              checked: table.getIsAllRowsSelected(),
-              onChange: table.getToggleAllRowsSelectedHandler(),
-            }}
-            aria-label="Select all the rows of the table"
```
ocshawn, line 62: ideally you would want something like "select all rows on page x of x"
and that would be enough

packages/portal-proto/src/features/GenomicTables/GenesTable/utils.tsx
```diff
@@ -50,30 +50,49 @@ export const useGenerateGenesTableColumns = ({
...
+              classNames={{
+                input: "checked:bg-accent checked:border-accent",
+              }}
+              aria-labelledby={`geneSelect-${row.original.symbol}`}
+              {...{
+                checked: row.getIsSelected(),
+                onChange: row.getToggleSelectedHandler(),
+              }}
+            />
+          </>
```
ocshawn: don't need to create a separate hidden label you should place an id on the name column and use aria-labelledby to point to that

packages/portal-proto/src/features/GenomicTables/SomaticMutationsTable/utils.tsx
ocshawn, line 1: same comments as packages/portal-proto/src/features/GenomicTables/GenesTable/utils.tsx

packages/portal-proto/src/features/cDave/Controls.tsx
```diff
@@ -144,8 +144,11 @@ const FieldControl: React.FC<FieldControlProps> = ({
             <label
               className="font-content font-medium text-md"
               htmlFor={`switch-${field.full}`}
+              aria-labelledby={`switch-${field.full}`}
             >
-              <Highlight highlight={searchTerm}>{displayName}</Highlight>
+              <span id={`switch-${field.full}`}>
+                <Highlight highlight={searchTerm}>{displayName}</Highlight>
+              </span>
```
ocshawn: revert this, old code looked to be more correct, dont know if you can target an id in a Switch this way though

ocshawn: line 204 has htmlFor={`switch-${field.full}`} i dont know if two elements can have the same target

packages/portal-proto/src/features/cases/CasesView/utils.tsx
ocshawn, line 1: same comments as packages/portal-proto/src/features/GenomicTables/GenesTable/utils.tsx

packages/portal-proto/src/features/facets/EnumFacet.tsx
```diff
@@ -442,6 +447,7 @@ const EnumFacet: React.FC<FacetCardProps<EnumFacetHooks>> = ({
                           >
                             <div className="flex-none">
                               <Checkbox
+                                aria-label="placeholder label in EnumFacet Component"
```
ocshawn: ??

packages/portal-proto/src/features/facets/EnumFacet.tsx
```diff
@@ -397,7 +403,6 @@ const EnumFacet: React.FC<FacetCardProps<EnumFacetHooks>> = ({
                                   e.currentTarget.checked,
                                 )
                               }
-                              aria-label={`checkbox for ${field}`}
```
ocshawn, line 400: don't need to create a separate hidden label you can use area-label, can just be aria-label={value}

packages/portal-proto/src/features/manageSets/ManageSetsTable.tsx
ocshawn, line 1: same comments as packages/portal-proto/src/features/GenomicTables/GenesTable/utils.tsx

packages/portal-proto/src/features/projectsCenter/ProjectsTable.tsx
ocshawn, line 1: same comments as packages/portal-proto/src/features/GenomicTables/GenesTable/utils.tsx

packages/portal-proto/src/features/set-operations/SetOperationTable.tsx
ocshawn, line 1: same comments as packages/portal-proto/src/features/GenomicTables/GenesTable/utils.tsx

ocshawn (changes requested):
some more notes
When on http://localhost:3000/v2/analysis_page?app=Projects for the main table when i use space-bar to select a checkbox it changes keyboard focus to somewhere else on the page, focus should stay on checkbox https://www.w3.org/WAI/ARIA/apg/patterns/checkbox/examples/checkbox/ 
this may be due to the hidden label way you were doing things and may be fixed when those issue are resolved

packages/portal-proto/src/components/Modals/SetModals/SavedSets.tsx
alilahrime: removed Select/Unselect text, also using aria-label directly now

packages/portal-proto/src/components/Modals/SetModals/SetTable.tsx
alilahrime: The checkbox component by default would always show the opposite value (unchecked if checked and vice versa and i thought this would mediate). reverted to using just aria-label tho

packages/portal-proto/src/features/GenomicTables/GenesTable/utils.tsx
alilahrime: added page and total page to be able to display it this way

packages/portal-proto/src/features/GenomicTables/GenesTable/utils.tsx
alilahrime: replaced w/ just aria-label like before

packages/portal-proto/src/features/GenomicTables/SomaticMutationsTable/utils.tsx
alilahrime: applied here too

packages/portal-proto/src/features/cDave/Controls.tsx
alilahrime: reverted the code, good catch thank you

packages/portal-proto/src/features/cases/CasesView/utils.tsx
alilahrime: applied same change

packages/portal-proto/src/features/facets/EnumFacet.tsx
alilahrime: changed to `Select the ${~value} facet`

packages/portal-proto/src/features/facets/EnumFacet.tsx
alilahrime: using just aria-label now

packages/portal-proto/src/features/manageSets/ManageSetsTable.tsx
alilahrime: added here too

packages/portal-proto/src/features/projectsCenter/ProjectsTable.tsx
alilahrime: done

packages/portal-proto/src/features/set-operations/SetOperationTable.tsx
alilahrime: done

packages/portal-proto/src/components/Modals/SetModals/SavedSets.tsx
```diff
@@ -78,17 +78,19 @@ const SavedSets: React.FC<SavedSetsProps> = ({
...
+              classNames={{
+                input: "checked:bg-accent checked:border-accent",
+              }}
+              aria-label={`Select ${row.original.setId}`}
+              {...{
+                checked: row.getIsSelected(),
+                onChange: row.getToggleSelectedHandler(),
+              }}
+            />
+          </>
```
ocshawn: does not need to be in `<></>` 

aria-label={`Select ${row.original.setId}`} should be whatever the user is selecting and no additional text, if the value is visible on the page it should be aria-labelledby instead and point to that value

packages/portal-proto/src/features/cases/CasesView/utils.tsx
ocshawn, line 1: see changes on packages/portal-proto/src/features/GenomicTables/SomaticMutationsTable/utils.tsx

Checkboxs dont need to be in `<></>` 
and
aria-label={`Select the ${row?.original?.project} case row`} should be whatever the user is selecting and no additional text, if the value is visible on the page it should be aria-labelledby instead and point to that value

ocshawn (commented):
have not gotten through everything yet but wanted to get this out

packages/portal-proto/src/features/cases/CasesView/utils.tsx
```diff
@@ -48,36 +48,42 @@ export const useGenerateCasesTableColumns = ({
...
-            classNames={{
-              input: "checked:bg-accent checked:border-accent",
-            }}
-            {...{
-              checked: table.getIsAllRowsSelected(),
-              onChange: table.getToggleAllRowsSelectedHandler(),
-            }}
-            aria-label="Select all the rows of the table"
-          />
+          <>
```
ocshawn: nitpic `<>` not needed

packages/portal-proto/src/features/facets/EnumFacet.tsx
```diff
@@ -388,6 +388,7 @@ const EnumFacet: React.FC<FacetCardProps<EnumFacetHooks>> = ({
                           <div className="flex-none">
                             <Checkbox
                               data-testid={`checkbox-${value}`}
+                              aria-label={`Select the ${value} facet`}
```
ocshawn: this can just be value or better to point to the actual value by using an ID

packages/portal-proto/src/features/facets/EnumFacet.tsx
```diff
@@ -442,6 +442,7 @@ const EnumFacet: React.FC<FacetCardProps<EnumFacetHooks>> = ({
                           >
                             <div className="flex-none">
                               <Checkbox
+                                aria-label={`Select the ${field} facet`}
```
ocshawn: what this this labeling? this is a loading view and should be disabled?

packages/portal-proto/src/features/manageSets/ManageSetsTable.tsx
```diff
@@ -206,23 +207,47 @@ const ManageSetsTable: React.FC<MangeSetsTableProps> = ({
...
-              input: "checked:bg-accent checked:border-accent",
-            }}
-            {...{
-              checked: table.getIsAllRowsSelected(),
-              onChange: table.getToggleAllRowsSelectedHandler(),
-            }}
-            data-testid="checkbox-select-all-sets"
-            aria-label="Select all the rows of the table"
-          />
+          <>
```
ocshawn: nitpick `<>` not needed

packages/portal-proto/src/features/projectsCenter/ProjectsTable.tsx
```diff
@@ -163,35 +163,40 @@ const ProjectsTable: React.FC = () => {
...
-            classNames={{
-              input: "checked:bg-accent checked:border-accent",
-            }}
-            {...{
-              checked: table.getIsAllRowsSelected(),
-              onChange: table.getToggleAllRowsSelectedHandler(),
-            }}
-            aria-label="Select all the rows of the table"
-          />
+          <>
```
ocshawn: nitpick `<>` not needed

packages/portal-proto/src/features/projectsCenter/ProjectsTable.tsx
```diff
@@ -163,35 +163,40 @@ const ProjectsTable: React.FC = () => {
...
-            classNames={{
-              input: "checked:bg-accent checked:border-accent",
-            }}
-            aria-label="checkbox for selecting table row"
-            {...{
-              checked: row.getIsSelected(),
-              onChange: row.getToggleSelectedHandler(),
-            }}
-          />
+          <>
```
ocshawn: nitpick `<>` not needed

packages/portal-proto/src/features/set-operations/SetOperationTable.tsx
```diff
@@ -86,23 +86,23 @@ export const SetOperationTable = ({
...
-            size="xs"
-            classNames={{
-              input: "checked:bg-accent checked:border-accent",
-            }}
-            value={row.original.operationKey}
-            aria-label="checkbox for selecting table row"
-            {...{
-              checked: selectedSets[row.original.operationKey],
-              onChange: (e) => {
+          <>
```
ocshawn: nitpick `<>` not needed

ocshawn (changes requested):
generally looking good, dont need to address the nitpicks

packages/portal-proto/src/features/facets/EnumFacet.tsx
alilahrime: removed it since its disabled/loading view. good catch

packages/portal-proto/src/features/facets/EnumFacet.tsx
alilahrime: do u have code suggestion for the "pointing" approach? otherwise i made the label just the value

ocshawn: thats fine, the pointing approach is to put an ID around the value and then do the aria-labelledBy like we did with the other ones

ocshawn (approved):
looks good to me

craigrbarnes (approved):
Changes look appropriate

PR #803, opened 2023-11-15
Qa 2076 manage sets create automation

bpierce01:
## Description
- Added tests for create set in manage set
- Added tests for some alternative paths, text and link checking
- Added Automation IDs for container that holds the bottom row of buttons inside a modal for custom filters, creating sets, adding genes to existing sets, etc. 

The manage set tests work will be continued in QA-2077, so this is not a finished product


packages/portal-proto/src/components/ClearStoreErrorBoundary.tsx
```diff
@@ -25,7 +25,7 @@ class ClearStoreErrorBoundary extends Component<
             An unexpected error has occurred. Your saved cohorts are being
             recovered, but your cart and sets may be lost.
           </p>
-          <ModalButtonContainer>
+          <ModalButtonContainer data-testid="modal-button-container">
```
uc-pperumala, line 28: If it's a button, this should start with `button-`

packages/portal-proto/src/components/InputEntityList/InputEntityList.tsx
```diff
@@ -269,7 +269,7 @@ const InputEntityList: React.FC<InputEntityListProps> = ({
           />
         )}
       </div>
-      <ButtonContainer>
+      <ButtonContainer data-testid="modal-button-container">
```
uc-pperumala, line 272: Same.
If it's a button, this should start with `button-`

uc-pperumala (approved):

.snyk
```diff
@@ -0,0 +1,2 @@
+language-settings:
```
bpierce01, line 1: Snyk uses python 3.7.4 by default, but that version of python does not support Pillow 10.0.1
So this .snyk file should update the version of python that Snyk uses to validate our dependencies 

Supporting document here: Document here: https://docs.snyk.io/scan-using-snyk/supported-languages-and-frameworks/python

packages/portal-proto/src/components/ClearStoreErrorBoundary.tsx
bpierce01: As discussed it is not a button, but rather the bottom portion of the modal that contains buttons

packages/portal-proto/src/components/InputEntityList/InputEntityList.tsx
bpierce01: As discussed it is not a button, but rather the bottom portion of the modal that contains buttons

.snyk
bpierce01: Had a conversation with the security team, and they said it is fine to try to see if it fixes the issue. We will know once it's merged in!

craigrbarnes (approved):

.snyk
craigrbarnes: I will merge now and we will see

PR #804, opened 2023-11-15
PEAR-1360 - New Cohort workflow based on max of 1 unsaved cohort

amlehman:
## Checklist

- [x] Added proper unit tests
      flagged issues


packages/core/src/coreCreateApi.ts
```diff
@@ -2,14 +2,19 @@ import {
...
 
-export const coreCreateApi = buildCreateApi(
-  coreModule(),
-  reactHooksModule({
-    useSelector: useCoreSelector as any,
-    useStore: useCoreStore as any,
-    useDispatch: useCoreDispatch,
-  }),
-);
+const coreCreateApi: CreateApi<keyof ApiModules<any, any, any, any>> =
```
amlehman, line 10: This fixes the typing on hooks created with `coreCreateApi`

packages/portal-proto/src/features/GenomicTables/SomaticMutationsTable/SMTableContainer.tsx
```diff
@@ -77,16 +76,14 @@ export interface SMTableContainerProps {
...
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  handleSurvivalPlotToggled = (_1: string, _2: string, _3: string) => null,
-
+  handleSurvivalPlotToggled = undefined,
   geneSymbol = undefined,
   projectId = undefined,
   genomicFilters = { mode: "and", root: {} },
   cohortFilters = { mode: "and", root: {} },
   caseFilter = undefined,
-  handleSsmToggled = () => null,
```
amlehman, line 88: These were creating a new instance when the component rerendering, causing the react table to rerender each time too.

wteouchicago:
Looking good.

## Issues
1. for uploading/importing the cohort (upload button on cohort bar), after clicking Save in the save cohort modal, the Import A New Cohort modal is still displayed. The Import...modal should be dismissed (just like current behaviour).

2. Looks like there's one location where cohorts should be saved and they are not: the 2nd and 3rd options in the Save New Cohort dropdown above the cases table and the cDAVE table. (The first option works as expected though.)
- you'll need to update the title of the modals to where you pick existing cohorts as well to "Save New Cohort..."
- and also show the Save Cohort modal instead of the Create Cohort one

___

Loving the little spinners on the buttons with +. So cute.

wteouchicago:
Issues 1-2 look good.

### New(?) Issues
3. Not sure if it's related to this branch or surfaces more due to changes here. I can file a separate ticket too. The spinner on the case counts in the cohort bar doesn't seem to go away after saving and switching to the new saved cohort. Steps:
- start with some cohort that has >0 cases; doesn't matter if it's saved or unsaved
- Cohort Bar > cases table. select some cases and use one of the options in Save New Cohort.
- save the cohort, then switch to it using the notification
- the spinner  on the cohort bar's case counts doesn't seem to ever go away; likewise the spinner on the files count in the expanded cohort bar.
- if you refresh the tab, the counts will show up. 

4. Mutation Frequency > the buttons in the genes and mutations tables no longer work to create/save cohorts. :(

amlehman:
Hi @wteouchicago,

3) Yeah, I'm think this is probably an existing/separate issue. 
4) Fixed

wteouchicago:
I'll file a separate ticket for 3.
4 looks good.

all good 👍

packages/core/src/coreCreateApi.ts
craigrbarnes: Nice

packages/portal-proto/src/components/Modals/SaveCohortModal.tsx
```diff
@@ -46,23 +51,59 @@ const SaveCohortModal = ({
...
+            coreDispatch(
+              setCohortMessage([`savedCohort|${newName}|${payload.id}`]),
+            );
+          } else {
+            coreDispatch(
+              setCohortMessage([
+                `savedCohortSetCurrent|${payload.name}|${payload.id}`,
+              ]),
+            );
+          }
```
craigrbarnes, line 104: Need to get the caseCount here as well, so add:
`coreDispatch(fetchCohortCaseCounts(payload.id));`

This will fix the spinner issue.

craigrbarnes (changes requested):
See my comment on SaveCohortModal.

craigrbarnes:
Changed ProteinPaint and OncoMatrix create cohort to use SaveCohort

siosonel:
> Changed ProteinPaint and OncoMatrix create cohort to use SaveCohort

@craigrbarnes Thanks for the updates and fixing the tests, I'll try to review tonight or by tomorrow morning.

wteouchicago:
spinner issue is fixed. 👍  I'll close the separate jira I filed.

siosonel:
@craigrbarnes I tested with qa-pink and qa-uat VPNs, and also without using the VPN, and I always see the following issues regardless:

- I'm getting a "There was a problem saving the cohort" in both the PP and OncoMatrix tools. This network error coincides with the error popup:
Screenshot 2023-11-21 at 10 58 03 PM

Also, the spinners are always spinning for me, the case counts does not seem to finish loading to get displayed. 
Screenshot 2023-11-21 at 10 58 24 PM

If it's urgent to merge these PR, it's okay to have the PP/OM tools to be temporarily broken, I can work on PP-specific issues on a separate branch.

wteouchicago:
Hi @siosonel -- I think saving cohorts require that you are running these steps: https://github.com/NCI-GDC/gdc-frontend-framework#running-auth-in-localhost-login-and-other-auth-related-actions
When Chrome gives you the error about the cert, type `thisisunsafe` - https://jasonmurray.org/posts/2021/thisisunsafe/

siosonel:
@wteouchicago I was running using ssl-proxy, using [a local https URL](https://localhost.gdc.cancer.gov:3010/v2/analysis_page?app=ProteinPaintApp), in Chrome and the cohort creation (using CreateCohortModal) worked without having to disable CORS restrictions. But this same setup is not working for in this branch. 

I also tried with Safari with disabled CORS and I still get the "There was a problem saving the cohort" error. My experience is that the full ssl + disabled CORS setup is only needed for features that require a user to be logged-in (such as Sequence Read), and not required for features like cohort creation that use public data.

I'll troubleshoot a bit more later today.

craigrbarnes:
> @wteouchicago I was running using ssl-proxy, using `localhost.gdc.cancer.gov` as hostname, in Chrome and the cohort creation (using CreateCohortModal) worked without having to disable CORS restrictions. But this same setup is not working for cohort creation in this branch.
> 
> I also tried with Safari with disabled CORS and I still get the "There was a problem saving the cohort" error. My experience is that the full ssl + disabled CORS setup is only needed for features that require a user to be logged-in (such as Sequence Read), and not required for features like cohort creation that use public data.
> 
> I'll troubleshoot a bit more later today.

I will also look into this. I would prefer to merge this PR without having to break PP/OM.

craigrbarnes:
@siosonel I do not see those errors; however, can you create a cohort from the project page and see if you get the same issue?

siosonel:
@craigrbarnes I got the same error from the project page, so the issue must be related to my dev setup, and not specific to the SaveCohortModal or PP/OM tools. The issue is likely related to not using the correct GDC API URL, `api.gdc.cancer.gov` versus `https://portal.gdc.cancer.gov/auth/v0`, plus maybe incorrect VPN? I tried the portal v0 API while on qa-pink VPN, and even but I get a Not Found Error. I even tried changing to `NEXT_PUBLIC_GDC_API=https://portal.gdc.cancer.gov/auth/v0` in `.env.development`. I'll keep looking into it.

Additional details:
In Safari (with the Disable CORS restrictions), the error for the same `https://portal.gdc.cancer.gov/auth/v0/cohorts?include_case_ids=true` endpoint is
```html
<!doctype html>
<html lang="en">
<head>
  <title>Not Found</title>
</head>
<body>
  <h1>Not Found</h1><p>The requested resource was not found on this server.</p>
</body>
</html>
```

siosonel:
@craigrbarnes It's likely that the `/cohorts?include_case_ids=true` endpoint error is **unrelated** to the save cohort issue that I'm seeing. In `https://portal.gdc.cancer.gov/v2/analysis_page?app=OncoMatrix` while on qa-pink VPN, I'm still able to create a cohort even while seeing that `/cohorts` error. 

I'll try to trace what triggers the "There was a problem saving the cohort" message in the GFF code.

wteouchicago:
@siosonel can you try with an incognito browser? Sometimes that error can happen if the context_id in your browser cannot be found on the backend.

siosonel:
> @siosonel can you try with an incognito browser? Sometimes that error can happen if the context_id in your browser cannot be found on the backend.

@wteouchicago Thanks, that fixed the errors! Verified to work in incognito mode for both Chrome and Safari. I'll approve the PR for now, but later there should not be an error even when not using incognito mode, most users will not be using that browser mode.

siosonel (approved):

wteouchicago:
@siosonel oh good that it worked! Craig/Phil will know this better, but I _think_ part of the issue is that we are switching environments (e.g. qa-uat, qa-pink) and a context_id stored when on one environment will not be available on another. Regular portal users shouldn't run into this situation.

craigrbarnes (approved):
The new code and changes are fine. Updated to the use of deep-compare are appreciated and will hope improve performance.

PR #808, opened 2023-11-16
Attempt to upgrade version of python SNYK tests with QA-2089

bpierce01:
## Description
- I updated the .snyk file, and put it in the directory with the holmes-py requirement file

It may not honor the fact we want Python version 3.9. If it does not, I will have to talk to security and come up with a new way to upgrade the version of python.

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -1,2 +1,0 @@
-language-settings:
-  python: '3.9'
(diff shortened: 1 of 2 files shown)


uc-pperumala (approved):

craigrbarnes (approved):
Look fine to me

PR #828, opened 2023-12-01
PEAR-1642 - Cohort Comparison - Clicking on "Open Venn Diagram" causes unexpected error and user is stuck in loop

amlehman:
## Description
\+ error boundary will redirect user to homepage if the store clear and refresh doesn't fix their issue rather than sticking them in an infinite loop 😅 

## Checklist

- [x] Added proper unit tests
      flagged issues


wteouchicago:
fix for crash works for me. I also tested the "redirect to home page" part by undoing the Set Operations fix and that worked too. 👍 👍

paribartandhakal (approved):
🥇

PR #839, opened 2023-12-06
fix(PEAR-1653): capitalize facet aria labels

alilahrime:
[pear-1653](https://gdc-ctds.atlassian.net/browse/PEAR-1653)

samples.days_to_collection
Screenshot 2023-12-06 at 2 28 22 PM
follow_ups.viral_hepatitis_serologies
Screenshot 2023-12-06 at 2 27 04 PM

### DIFF
--- a/packages/portal-proto/src/features/facets/EnumFacet.tsx
+++ b/packages/portal-proto/src/features/facets/EnumFacet.tsx
@@ -296,7 +296,7 @@ const EnumFacet: React.FC<FacetCardProps<EnumFacetHooks>> = ({
                     clearFilters(field);
                     dismissCallback(field);
                   }}
-                  aria-label="remove the facet"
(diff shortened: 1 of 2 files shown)


wteouchicago:
Do we want to capitalize here or lowercase? Which do we normally do for aria-labels? I saw a bunch of lowercase ones at a glance, but maybe there are capitalized ones elsewhere.
any ideas - @amlehman / @paribartandhakal / @cemile-ctds ?
@bpierce01 -- you caught this via a test, right? what did the test expect?

cemile-ctds:
It doesn't really matter if it's sentence case, camel case, or all lowercase – most screen readers will pronounce it the same way. However, it's best to avoid using all caps for the label because that can mess with the screen reader's voice inflection and might cause some words to be read as separate characters.

PR #841, opened 2023-12-06
fix(PEAR-1622): cohort bar "Save New Cohort" dropdown Tooltip should not be displayed

alilahrime:
[pear-1622](https://gdc-ctds.atlassian.net/browse/PEAR-1622)


wteouchicago (changes requested):
The other buttons above the table now show the tooltip intended for the Save New Cohort button. 
Screenshot 2023-12-06 at 5 29 03 PM

alilahrime:
> The other buttons above the table now show the tooltip intended for the Save New Cohort button. Screenshot 2023-12-06 at 5 29 03 PM

tooltip shouldn't show up for the other drop downs now. also shouldn't show when its disabled (before selecting at least 1 case). i think this is a good for consistency since it'd be inaccessible a tooltip if navigating via keyboard

wteouchicago:
I verified that the tooltip is no longer on the other buttons. 

I also verified that if you are hovering over a menu option, the tooltip on Save New Cohort doesn't show up.

But I'm not sure we want to hide the tooltip when the button is disabled. This is because the tooltip was also to help people understand that they need to select something so as to use the button. Also, there is another Save New Cohort button that has a tooltip when disabled -- Projects table.

@cemile-ctds do you know if it's better to have no tooltip when something is disabled (and nobody can get the information) or is it better to have a tooltip when something is disabled (so some users can get the information)? 

btw, there are other places in the portal that has tooltips on a disabled option/button, e.g. Set Operations selection page when a row contain 0 entities.

wteouchicago:
synced with @cemile-ctds 

@alilahrime  -- please keep the tooltip displayed on the disabled button. 

in the future, we'll make tooltips on disabled buttons across the whole portal more informative (separate ticket)

alilahrime:
> synced with @cemile-ctds
> 
> @alilahrime -- please keep the tooltip displayed on the disabled button.
> 
> in the future, we'll make tooltips on disabled buttons across the whole portal more informative (separate ticket)

got it, added tooltip back back for disabled cases cohort button and it goes away when menu is open. 

okay sounds good for future buttons 👍🏽

wteouchicago (commented):
works for me. 👍

wteouchicago (approved):

craigrbarnes (approved):

PR #848, opened 2023-12-07
PEAR-1628 - Analysis Center - 508 Compliance issues

amlehman:
## Checklist

      flagged issues


packages/portal-proto/src/features/layout/UserFlowVariedPages.tsx
```diff
@@ -181,14 +181,14 @@ export const UserFlowVariedPages = ({
         <Header {...{ headerElements, indexPath, Options }} />
       </header>
       <ClearStoreErrorBoundary>
-        <div
+        <aside
```
amlehman, line 184: I think `<aside>` is a good fit for the cohort management bar, from [here](https://www.w3.org/TR/2011/WD-html5-author-20110809/the-aside-element.html):

> The aside element represents a section of a page that consists of content that is tangentially related to the content around the aside element, and which could be considered separate from that content

but we could also just have this as a `<region>` element if we think it doesn't fit enough.

paribartandhakal (approved):

PR #857, opened 2023-12-12
PEAR-1585 - Enforce max of 1 unsaved cohort portal wide

amlehman:
## Description
- Changes methods where you can add a cohort to deal with either a saved or unsaved cohort (but not both) and renames them to make that clear
- Throws an exception if an unsaved cohort is created when one already exists

Easiest way to test it is to comment out the disabled condition on the add cohort button [here](https://github.com/NCI-GDC/gdc-frontend-framework/blob/develop/packages/portal-proto/src/features/cohortBuilder/CohortManager.tsx#L523)

## Checklist

- [x] Added proper unit tests
      flagged issues


craigrbarnes (approved):
Changes are fine

PR #859, opened 2023-12-12
PEAR-1495: Add tooltip to buttons on query expressions

paribartandhakal:
## Checklist
- [ ] Added proper unit tests
      flagged issues

https://github.com/NCI-GDC/gdc-frontend-framework/assets/101295912/02e0cb35-1e6f-4025-8629-83ba4f854c15

### DIFF
--- a/packages/portal-proto/src/features/cohortBuilder/QueryExpressionSection.tsx
+++ b/packages/portal-proto/src/features/cohortBuilder/QueryExpressionSection.tsx
@@ -10,6 +10,7 @@ import { omit } from "lodash";
 import { useCoreDispatch, clearCohortFilters, FilterSet } from "@gff/core";
 import OverflowTooltippedLabel from "@/components/OverflowTooltippedLabel";
 import { convertFilterToComponent } from "./QueryRepresentation";
+import { Tooltip } from "@mantine/core";
 
 const QueryExpressionContainer = tw.div`
   flex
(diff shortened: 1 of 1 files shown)


wteouchicago:
still looks good 👍

cemile-ctds:
1. When the cohort query expressions section is in the empty state, the tooltip for the expand/collapse button doesn't appear. Also, when the cohort query expressions section is in the empty state, the expand/collapse button should also be in the inactive state. @wteouchicago 

2. The "Display Rows" button doesn't have active state colors. Please refer to the button states for guidance:
states

wteouchicago:
for 1) yeah, in a later ticket, i think we should go back to what we had before where there was a placeholder message about how you are viewing all of the gdc and that you should use the cohort builder to start filtering.
- if it's an easy fix to disable the button when there are no filters on the cohort, go ahead, @paribartandhakal  -- otherwise, i'll file a follow up. 
- also, if it's easy to add a tooltip when it's disabled, we can go with "No values to expand/collapse". otherwise, i'll put it in the follow-up ticket

for 2) again, if it's a quick fix, go ahead. else, i'll put in a follow-up.

(basically I'm trying to prioritise the main tooltips part for our soft launch)

wteouchicago:
filing a follow-up. pari's working on base path things

amlehman (approved):

PR #871, opened 2023-12-19
PEAR-762 - Put context id in local storage

amlehman:
## Checklist

- [x] Added proper unit tests
      flagged issues

### DIFF
--- a/packages/core/src/listeners.ts
+++ b/packages/core/src/listeners.ts
@@ -118,7 +118,7 @@ startCoreListening({
     if (!Cookies.get("gdc_context_id")) {
       const contextId = localStorage.getItem("gdc_context_id");
       if (contextId) {
-        Cookies.set("gdc_context_id", contextId);
+        Cookies.set("gdc_context_id", contextId, { domain: ".gdc.cancer.gov" });
       }
     }
   },
--- a/packages/core/src/listeners.ts
+++ b/packages/core/src/listeners.ts
@@ -4,6 +4,7 @@ import {
   isFulfilled,
 } from "@reduxjs/toolkit";
 import type { TypedStartListening } from "@reduxjs/toolkit";
+import Cookies from "js-cookie";
 import { CoreDispatch } from "./store";
 import { CoreState } from "./reducers";
 import {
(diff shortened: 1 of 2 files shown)


packages/core/src/listeners.ts
```diff
@@ -85,3 +87,39 @@ startCoreListening({
...
+    cohortApiSlice.endpoints.getCohortsByContextId.matchPending,
+    cohortApiSlice.endpoints.getCohortById.matchPending,
+    cohortApiSlice.endpoints.addCohort.matchPending,
+    cohortApiSlice.endpoints.updateCohort.matchPending,
+    cohortApiSlice.endpoints.deleteCohort.matchPending,
+  ),
+  effect: async () => {
+    // If cookie has been deleted, restore it from local storage
+    if (!Cookies.get("gdc_context_id")) {
+      const contextId = localStorage.getItem("gdc_context_id");
```
paribartandhakal, line 119: Will there be a case when both the localStorage and Cookies wont have this? If so what happens.

amlehman: I believe the backend will create a context id and attach the set-cookie header if the request doesn't have a context id.

paribartandhakal (dismissed):
🥇

wteouchicago:
I want to understand something / something doesn't feel right:
- incognito
- verified localstorage and cookie doesn't have the context_id
- make a cohort, save, e.g. "Eye"
- verified localstorage and cookie now have the context_id
- delete the context_id from cookie, then nuked the whole cookie for good measure
- tried saving -- ran into cohort saving error
- tried refreshing -- lost my "Eye" cohort
- but the cookie now has the context_id

I would have thought that now that the cookie has the context_id, I should be able to retrieve my "Eye" cohort that was saved and also be able to modify "Eye".

I then made another cohort and saved that. Now I have 2 context_ids in the cookie. The "old"/"first" one has an expiration/max-age set to "Session".

amlehman:
Hi @wteouchicago, I see the issue you're describing. Not sure how it was working for me before but I've pushed a change that should fix the cookie not getting added to the requests. Could you test again when you get a chance?

wteouchicago:
very nice -- everything is working as expected and smoothly. 👍 I tried variations of deleting cookies, deleting the context_id from cookie or from localstorage and nothing was disrupted.

craigrbarnes (approved):
Works for me as well

PR #891, opened 2024-01-03
Restored and added data-testids QA-2122

bpierce01:
## Description
- Added IDs for core tool navigation
- Restored ID for manage sets header

### DIFF
--- a/packages/portal-proto/src/features/layout/Header.tsx
+++ b/packages/portal-proto/src/features/layout/Header.tsx
@@ -151,7 +151,7 @@ export const Header: React.FC<HeaderProps> = ({
           </a>
           <Link href="/manage_sets" passHref>
             <a
-              data-testid="link-header-manage-sets"
+              data-testid="button-header-manage-sets"
               className={`p-1 rounded-md ${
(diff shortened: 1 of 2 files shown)


packages/portal-proto/src/features/layout/Header.tsx
```diff
@@ -151,7 +151,7 @@ export const Header: React.FC<HeaderProps> = ({
           </a>
           <Link href="/manage_sets" passHref>
             <a
-              data-testid="link-header-manage-sets"
+              data-testid="button-header-manage-sets"
```
paribartandhakal, line 154: I know that you have it as such in the automation but is it okay to change it there to make it consistent as this is now a link?

bpierce01: The items on the headers need to be all buttons or all links, they cannot be both. We have things labeled as buttons but not as links, so this would be more consistent in our repo

paribartandhakal (approved):

ocshawn (approved):
looks good, sorry for changing it in the first place

craigrbarnes (approved):

PR #904, opened 2024-01-09
PEAR-1713 - Combination of Save-As and Replace results in changes being lost

amlehman:
## Description
Changes from the first cohort were getting discarded before the second cohort was saved with the combination of save as + replace. Most of the changes are just reorganizing the saving logic so its easier to follow. 

## Checklist

- [x] Added proper unit tests
      flagged issues


wteouchicago:
Worked as expected in various combinations of Save/Save-As/Replace and with and without unsaved changes. 👍

paribartandhakal (approved):

PR #907, opened 2024-01-09
adding smoke-test tags for Data Portal v2 automation testing

AmyCMP:
## Description
while testing use this command `gauge run --tags "smoke-test" specs`

### DIFF
--- a/holmes-py/specs/gdc_data_portal_v2/analysis_center/navigation.spec
+++ b/holmes-py/specs/gdc_data_portal_v2/analysis_center/navigation.spec
@@ -5,7 +5,7 @@ Owner		    : GDC QA
 Description		: Test analysis center navigation
 Test-Case       : N/A
 
-tags: gdc-data-portal-v2, navigation, analysis-center
+tags: gdc-data-portal-v2, navigation, analysis-center, smoke-test
 
 ## Navigate to the analysis center
 * On GDC Data Portal V2 app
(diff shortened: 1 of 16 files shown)


bpierce01:
Where you marked spec files to be a smoke test, I think we only need 1 test per folder. Thoughts @uc-pperumala ?

uc-pperumala:
Thanks, @AmyCMP! It looks good. However, I had a question. @bpierce01 this may apply more to you. I am wondering why are we placing `smoke-test` at the file level and not at the scenario-level? 

Technically, smoke tests should just be handful of test scenarios.

bpierce01:
> Thanks, @AmyCMP! It looks good. However, I had a question. @bpierce01 this may apply more to you. I am wondering why are we placing `smoke-test` at the file level and not at the scenario-level?
> 
> Technically, smoke tests should just be handful of test scenarios.

My thought it should be on the spec level for two reasons:
1. The spec files are already small enough to where it's the building block of a test
2. Some scenarios are dependent on pre-existing scenarios having been executed 

Overall, suggestion is for the PR is to reduce the number of spec files that have the smoke test tag

bpierce01 (changes requested):
reduce the number of spec files that have the smoke test tag

AmyCMP:
> Thanks, @AmyCMP! It looks good. However, I had a question. @bpierce01 this may apply more to you. I am wondering why are we placing `smoke-test` at the file level and not at the scenario-level?
> 
> Technically, smoke tests should just be handful of test scenarios.



> reduce the number of spec files that have the smoke test tag

The number of spec files with smoke test tags have been reduced..

bpierce01 (approved):

PR #917, opened 2024-01-12
PEAR-1731: fix spinning case counts in save -> change cohort -> saveAs

craigrbarnes:
## Description
Fixes the issue reported in PEAR-1731. It also should ensure that case counts are not affected by changing the cohort and then switching to another before the fetch is fulfilled, as it now passes the cohort when the cohort changes. 
## Checklist

- [ ] Added proper unit tests
      flagged issues


packages/core/src/features/cohort/availableCohortsSlice.ts
```diff
@@ -1001,7 +1001,7 @@ const slice = createSlice({
           cohortsAdapter.updateOne(state, {
             id: action.meta.arg ?? getCurrentCohort(state),
             changes: {
-              counts: { ...NullCountsData, status: "pending" },
+              counts: { ...NullCountsData, status: "rejected" },
```
craigrbarnes, line 1005: If there is an error, set it to rejected.  In the future, we can either log or visually indicate an error. Even if we don't and should result in "pending"

amlehman (approved):

wteouchicago:
This worked for me to fix the original reported bug and also the two new ones Brent noted that involves infinitely spinning spinners on the case counts. 👍

PR #928, opened 2024-01-18
PEAR-1731,1743: Fix Case Counts race condition and infinite spinning 

craigrbarnes:
## Description
Fixes a race condition with caseCounts. fetchCaseCounts now requires a cohortId as a parameter and passes the filters to the response. Upon fulfillment, if the query filters does not match the current value of the cohort filters, the count values are ignored.

Fixes an issue with removeCohort and caseCounts

Specialized the listeners for new cohorts, removed cohorts, and cohort changes.
## Checklist

- [ ] Added proper unit tests
      flagged issues

### DIFF
--- a/packages/core/src/features/cohort/availableCohortsSlice.ts
+++ b/packages/core/src/features/cohort/availableCohortsSlice.ts
@@ -1015,7 +1015,6 @@ const slice = createSlice({
         if (!isEqual(cohort?.filters, response.cohortFilters)) {
           // count request is not for the current cohort
           // TODO: possibly add logging to track this
-          console.error("fetchCohortCaseCounts: cohort filters do not match");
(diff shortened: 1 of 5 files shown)


wteouchicago:
This worked for me. 👍 
I tested these two areas:
- PEAR-1731 -- endless spinners. 
- PEAR-1743 -- race condition. I tried with combos of adding/removing filters fast, changing cohorts, discarding unsaved cohort changes and didn't run into any unusual counts or counts replacing each other. That said, it may be easier to trigger race conditions on a qa server since running locally can be slow and I may be forced to click more slowly. 

Once code review passes, I recommend we put this on qa-orange for more testing of this area and to see if we catch oddities with local filters (or adding ssm or gene to cohort from MF).

amlehman (approved):

PR #931, opened 2024-01-18
PEAR-1749: Add a custom 404 page

paribartandhakal:
## Checklist
      flagged issues

https://github.com/NCI-GDC/gdc-frontend-framework/assets/101295912/01cce457-1cb5-4469-a7d4-0fb4f952b82a


packages/portal-proto/src/pages/404.tsx
```diff
@@ -0,0 +1,31 @@
...
+        <title>Page Not Found</title>
+        <meta
+          property="og:title"
+          content="GDC Data Portal Page Not Found"
+          key="gdc-page-not-found"
+        />
+      </Head>
+      <div className="flex flex-col justify-center items-center h-full p-8 gap-2">
+        <h1 className="text-2xl">Page Not Found</h1>
+        <h2>Sorry, we couldn&apos;t find the page you were looking for.</h2>
```
paribartandhakal: Got `Page should contain a level-one heading` warning from the axe devtools otherwise.

craigrbarnes: Looking at [https://dequeuniversity.com/rules/axe/4.1/page-has-heading-one](https://dequeuniversity.com/rules/axe/4.1/page-has-heading-one) wonder if adding role="heading" aria-level="1" would be appropriate.

paribartandhakal: This is kind of similar to setting the first one as `h1`. However, the problem is that if don't make the next line as `h2`, then I get `Heading levels should only increase by one`. This is because the footer contains `h3` tag.
Screenshot 2024-01-18 at 2 41 19 PM

wteouchicago:
looks good. very nice! 👍 
pinging @cemile-ctds in case she has opinions about the 508 thing.

packages/portal-proto/src/pages/404.tsx
cemile-ctds: Starting with an `<h1>` for headings makes sense, but it's not required by WCAG or HTML 5.2. While having a logical heading hierarchy is recommended, not having an `<h1>` doesn't violate WCAG.

However, it's not ideal to skip heading levels (like going from `<h2>` to `<h4>`). This may make the structure less clear, but it's not a WCAG failure.

A WCAG failure occurs when text that looks like a heading isn't marked as one, or vice versa. WCAG emphasizes using the most appropriate structure for conveying information, including headings. While it doesn't force you to use headings, if you visually style text to indicate a section, it should be marked up as a heading.

Additionally, WCAG insists on a logical order for headings. While skipping levels is allowed, heading levels should align with the content's hierarchy. In simpler terms, less important headings should have lower hierarchy levels.

In our case, what we're not aligning with is a logical heading order. Also, we incorrectly labeled some text as a heading when it wasn't one.

To address these issues, refrain from marking "National Cancer Institute at the National Institutes of Health" as a heading. Additionally, ensure that the titles of the footer navigation are at the `<h2>` level.

Screenshot 2024-01-19 at 14 21 57

craigrbarnes: I would suggest fixing these in this PR. It does increase the scope, but we might as well fix it now.

paribartandhakal: I have done as discussed. Please verify it once.

cemile-ctds:
Looks good on my side.

craigrbarnes (approved):

PR #933, opened 2024-01-18
PEAR-1752: add requestID to most slices that need it.

craigrbarnes:
## Description
Adds `requestId` to all slices to prevent race conditions.

## Checklist

- [ ] Added proper unit tests
      flagged issues


craigrbarnes:
Code-wise lots of repetition to review. Make sure things are consistent and I am not missing any slice.

packages/core/src/features/annotations/annotationsSlice.ts
```diff
@@ -42,17 +43,19 @@ const slice = createSlice({
   extraReducers: (builder) => {
     builder
       .addCase(fetchAnnotations.fulfilled, (state, action) => {
+        if (state.requestId != action.meta.requestId) return state;
```
philtom-ctds, line 46: (This comment is not specific to this slice. Just placing it on the first one to allow for replies)

I see two potential issue with this pattern:
1. If any page makes multiple of the same request but with different request parameters, then only the last request will win.  Each reducer needs to be reviewed independently to see what it's doing.  If it completely replaces the state, then it should be ok.  If it's updated part of the state, especially if it's keyed based on the response, then it'll likely be problematic.
2. If the slice state is persisted to local storage, then we have the potential for any portal v2 tab to block the activity of any other portal v2 tab. While we do not officially support multiple tabs, this pattern may exacerbate any existing issues.

craigrbarnes: My understanding is we always want the last request to win. All reducers in the PR except for two, which I will revert, completely replaces the state. 

None of these are persisted into local storage.

craigrbarnes: I reverted GeneSymbol, which we should replace with a static file (tech debt), and ImageViewer, as they both do partial updates to the state.

craigrbarnes: GeneSymbol has no issue with a race condition as it will replace a gene symbol with the same value. Which is why it should not need a slice or even be in the state.

packages/core/src/features/annotations/annotationsSlice.ts
```diff
@@ -42,17 +43,19 @@ const slice = createSlice({
...
       })
-      .addCase(fetchAnnotations.pending, (state) => {
+      .addCase(fetchAnnotations.pending, (state, action) => {
         state.status = "pending";
+        state.requestId = action.meta.requestId;
         state.error = undefined;
-
         return state;
       })
       .addCase(fetchAnnotations.rejected, (state) => {
```
philtom-ctds: It seems like the `rejected` case needs to check the `requestId` as well.  Otherwise, in the normal case when responses are handled in the same order as the requests, a rejection from an earlier request will cause the application think the latest request had an error.

This applies to ever reducer that is using this `requestId` pattern to prevent race conditions.

packages/core/src/features/cancerDistribution/cnvPlot.ts
```diff
@@ -230,8 +231,11 @@ const slice = createSlice({
           state = castDraft(initialState);
           state.status = "rejected";
           state.error = response.errors.message;
+          return state;
         }
 
+        if (state.requestId != action.meta.requestId) return state;
```
philtom-ctds: I think this check needs to be the first statement of the case.  As it currently is, an error from an earlier response would be made available to the application while we're waiting for the latest request to resolve.  See my earlier comment in the annotations slice about handling the `rejected` case.  I think the same reasoning applies to handling errors in fulfilled requested.

this comment applies to every reducer that is using requestId to process only the latest request and is handling errors within the fulfilled case.

packages/core/src/features/genomic/ssmsSummary/ssmsSummarySlice.ts
```diff
@@ -62,6 +63,8 @@ const slice = createSlice({
   extraReducers: (builder) => {
     builder
       .addCase(fetchSsms.fulfilled, (state, action) => {
+        if (state.requestId != action.meta.requestId) return state;
+
         const response = action.payload;
 
         state.summaryData = response.data.hits.map((hit) => ({
```
philtom-ctds, line 69: state.summaryData is set here.  state.ssms is not.  is state.ssms unused, which makes this change ok.  Or is it set elsewhere, which would require additional logic for preventing race-conditions?

packages/core/src/features/annotations/annotationsSlice.ts
philtom-ctds: I took a look at the image viewer reducer, and I think there's a bug there.  Since it's doing an object spread of the existing data and the new data, it will accumulate key-value entries and never remove entries.  If I'm reading the code correctly, then it's a bug separate from this race-condition fix.  However, depending on how the bug is resolved, this race condition fix may need to be applied.

philtom-ctds (approved):

PR #934, opened 2024-01-19
PEAR-1702 - Mutation Frequency data not updating sometimes

amlehman:
## Description
Clear internal case set when removing gene/ssm filter from cohort otherwise the cohort will be "stuck" with the ssm/gene filter 

## Checklist

- [x] Added proper unit tests
      flagged issues

### DIFF
--- a/packages/core/src/features/cohort/availableCohortsSlice.ts
+++ b/packages/core/src/features/cohort/availableCohortsSlice.ts
@@ -752,7 +752,6 @@ const slice = createSlice({
         });
     },
     removeCohortFilter: (state, action: PayloadAction<string>) => {
-      // todo clear case set if not needed
       const root = state.entities[getCurrentCohort(state)]?.filters.root;
       if (!root) {
         return;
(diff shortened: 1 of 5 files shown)


wteouchicago:
I'm seeing the data not update sometimes and I think I've narrowed it down a bit. Maybe.

Steps:
- lung cohort from home page > Mutation Frequency
- Mutations tab, add top SSM (KRAS G12C) to cohort
- As expected: KRAS G12C is still the top SSM as expected, survival plot doesn't plot; Genes tab has KRAS as top gene
- Mutations tab: remove KRAS G12C from cohort using the toggle. wait for everything to finish loading
- now add the 2nd top SSM to cohort: EGFR L858R
- **note that the mutations table doesn't seem to update properly. EGFR L858R should be the top SSM, but it's not in the table (at least not the first page)**
- go to Genes tab and note that EGFR is the top gene, as expected.

amlehman:
Hi @wteouchicago, I think I've fixed the issue you were seeing (slightly different cause, the request framework we're using caches requests by argument parameters. Because we are reusing the internal case set id when we update filters, those requests were not properly getting cache expired). Can you test again when you get the chance?

wteouchicago:
looks good :) 👍

paribartandhakal (approved):

craigrbarnes (approved):

PR #936, opened 2024-01-22
PEAR-1762: Clean up image viewer/detail slice

paribartandhakal:
## Description
1. The state will not constantly grow as the state is being reset when the component unmounts.

2. It was found that the chance or scenario of race condition is highly unlikely. That might be a different case once Image Viewer will be an app.

**Note:**
Please verify that the multiple image viewer works as before.

## Checklist
- [ ] Added proper unit tests
      flagged issues


wteouchicago:
looks ok to me -- didn't spot anything unusual about multiple images viewing. 👍

craigrbarnes (approved):
Good solution to ensure the state is cleared.

PR #945, opened 2024-01-26
PEAR-1701 - Mutation Frequency, mutations survival plot flicker, slowness

amlehman:
## Description
Refactor to:
- Group all the top gene/ssms logic together in a hook so its easier to understand 
- Rewrite useEffects so they don't have any ignored dependencies and are not triggering each other to prevent the infinite render bugs we keep seeing
- Switch topGeneSSMSSlice to use createApi. Our slices return a new instance of the data every render which makes it hard to correctly use useEffect

## Checklist

- [x] Added proper unit tests
      flagged issues


wteouchicago:
Looks good. I checked that there are no longer flickers nor performance issues, whether at 0 cases or >0 cases. I also did some spot checks of the data displayed and they looked fine. 👍

paribartandhakal (approved):
🥇

PR #971, opened 2024-02-08
PEAR-1765 - Focus doesn't move to the toast message

amlehman:
## Description
Cemile and I chatted about this ticket, did some research, and came to the conclusion that moving focus to the toast message was not the right pattern (https://adrianroselli.com/2020/01/defining-toast-messages.html#RoleAlert). So now this ticket just fixes an issue where toast messages were not getting read by the screenreader in analysis center apps.  

\+ Also addresses PEAR-1726 - Keyboard focus not returning to the active element (table button element) after Summary Modal closes.

## Checklist

- [ ] Added proper unit tests
      flagged issues


packages/portal-proto/src/components/AccessibleNotificationWrapper.tsx
```diff
@@ -0,0 +1,34 @@
...
+    ref?.current?.focus();
+
+    return () => {
+      window.setTimeout(returnFocus, 10);
+    };
+  }, []);
+
+  return (
+    // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
+    <div aria-live="assertive" ref={ref} tabIndex={0}>
```
amlehman: This feels a little off to me, not sure if there's a better way to move focus to the modal?

cemile-ctds:
The focus moves to the toast message as expected, and the screen reader reads it out. Additionally, the focus returns to the selected button.

However, the screen reader reads the toast message twice. Since the time interval is not long enough, we only hear part of the second read. I tried to figure out why this was happening and how we could fix it, but I couldn't find a solution. I suggest approving it as is for now. After we push this to production, I'd like to test it with NVDA using Assistiv Labs.

Screenshot 2024-02-12 at 15 59 13

amlehman:
HI @cemile-ctds, I looked into this and the root cause is the fact that the mantine notifications have `role=alert` on them. That is a correct/appropriate role for a notification but it does seem to have a known problem of screenreaders reading content of a focused alert twice (see https://webaim.org/blog/aria-cause-solution under "role=alert"). Not sure what the best fix here is 🤔

cemile-ctds:
The screen reader reads the toast message, and it works fine on my end.

I'll be opening a new ticket for the 508 violations related to the actionable elements in the toast message.

packages/portal-proto/src/components/CollapsibleContainer.tsx
```diff
@@ -83,13 +87,20 @@ export const CollapsibleContainer = (
...
         </div>
       </div>
-      <Collapse
-        in={!isCollapsed}
-        transitionDuration={200}
-        transitionTimingFunction="linear"
+      <div
+        aria-hidden={isCollapsed}
+        style={{ height: !isCollapsed ? descHeight : 0 }}
+        className="transition-[height] duration-300 overflow-hidden"
```
paribartandhakal, line 93: @amlehman  I am wondering what `transition-[height]` does?

paribartandhakal (approved):

packages/portal-proto/src/components/CollapsibleContainer.tsx
amlehman: It's the css transition property for this transition: https://developer.mozilla.org/en-US/docs/Web/CSS/transition-property. We have it specified as a custom property in the tailwind config here: https://github.com/NCI-GDC/gdc-frontend-framework/blob/develop/packages/portal-proto/tailwind.config.js#L845

PR #977, opened 2024-02-15
QA-2074 Check for gdc_context_id cookie is accessible using Javascript

uc-pperumala:
## Description
Check whether the `gdc_context_id` is accessible using Javascript

Additional check was made to ensure that the `gdc_context_id` was generated using uuid4.


bpierce01 (approved):
Looks good, test passes: approving

PR #979, opened 2024-02-16
(PEAR-1764): properly use mantine Switch component built in label

ocshawn:
## Description

- change instances of mantine Switch component to use built in label in preparation for mantine patch 
- convert SwitchSpring to use mantine Switch component
- adjust tests and test ids to better match new SwitchSpring code

## Checklist

- [x] Added proper unit tests
      flagged issues

### DIFF
--- a/packages/portal-proto/src/components/SwitchSpring/shared/SwitchSpring.tsx
+++ b/packages/portal-proto/src/components/SwitchSpring/shared/SwitchSpring.tsx
@@ -1,8 +1,5 @@
 import React, { ReactNode } from "react";
-import { useSpring, animated } from "@react-spring/web";
-import { Tooltip } from "@mantine/core";
-import classNames from "classnames";
-import { createKeyboardAccessibleFunction } from "src/utils/";
+import { Tooltip, Switch } from "@mantine/core";
 
 interface SwitchSpringProps {
   isActive: boolean;
@@ -27,22 +24,9 @@ const SwitchSpring: React.FC<SwitchSpringProps> = ({
   margin,
   survivalProps,
 }: SwitchSpringProps) => {
-  const ballSpring = useSpring({
-    width: 20,
-    y: 0,
-    x: isActive ? 12 : -4,
-  });
-
-  const sliderSpring = useSpring({
-    width: 35,
-  });
-
   const { plot } = survivalProps ?? { plot: "" };
 
   const toggleSwitch = () => {
-    if (disabled) {
-      return;
-    }
     // todo: if used for > 2 icons refactor to use switch(icon) statement
     icon
       ? handleSwitch(selected[`symbol`], selected[`name`], plot)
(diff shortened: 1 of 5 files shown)


ocshawn:
the issues found in wave will be fixed when we update mantine

packages/portal-proto/src/components/SwitchSpring/shared/SwitchSpring.tsx
```diff
@@ -55,36 +39,27 @@ const SwitchSpring: React.FC<SwitchSpringProps> = ({
...
-            isActive && `bg-lightgray`
-          } h-5`}
-        >
-          <div className={margin}>{icon}</div>
-        </animated.div>
-      </animated.div>
+      {/* div added otherwise tooltip does not work in mantine 6 */}
+      <div>
+        <Switch
+          data-testid="switchSpring-checkbox"
```
bpierce01: Can you call this "button-switchspring" instead?

bpierce01 (changes requested):
Change the "switchSpring-checkbox" to "button-switchspring". Thank you!

packages/portal-proto/src/components/SwitchSpring/shared/SwitchSpring.tsx
ocshawn: yes we can, just FYI the label is applied to the checkbox when the component is rendered not to the button

bpierce01 (approved):

craigrbarnes:
One issue I noticed is that the original switch has a round control:

image
But this does not:
image

This should be fixed, unless @wteouchicago is ok with it (The documentation  videos have the square buttons)
Additionally, the new control is a bit wider than the original, not sure that is an issue.

craigrbarnes (changes requested):
The switch control is now square, it should be round.

wteouchicago:
Round does seem more ideal (and will match the cancer gene census toggle), but if making it round is a pain, let me know. Also checking @cemile-ctds 's opinion.

Edit: mostly I would prefer the background portion to be the same type of shape as the switch in the foreground.

cemile-ctds:
@craigrbarnes @wteouchicago 
The initial design of the switch buttons on the tables was square. The main reason for this decision is to differentiate these buttons from the on/off buttons on the filters and to customize the column list. Please refer to the table mockup link below and the reference screenshots for more details.

If we are going to follow this design decision, we should also ensure that the background part matches the same shape as the foreground part. 

However, if this is not something we can implement, then please make the switch buttons round; otherwise, please keep the design as it is.

Screenshot 2024-02-22 at 10 51 48


Table mockups: https://xd.adobe.com/view/9494a67c-0198-434e-a925-ac008f31c9b4-c1c7/grid

ocshawn:
@cemile-ctds updated styling to have square background please re evaluate

cemile-ctds:
Screenshot 2024-03-05 at 11 58 10
Look really good @ocshawn. Thank you so much.

amlehman (approved):

craigrbarnes (approved):
Looks good!

PR #980, opened 2024-02-16
DEV 2513 amazon linux

qqiao2024:
## Description
Update docker image to use amazon linux

## Checklist

- [ ] Added proper unit tests
      flagged issues


.gitlab-ci.yml
```diff
@@ -82,6 +82,11 @@ Build and push image:
       docker build \
       --build-arg "NPM_REGISTRY=https://nexus.osdc.io/repository/npm-all/" \
       --build-arg "BUILD_SHORT_SHA=$CI_COMMIT_SHORT_SHA" \
+      --build-arg "BASE_VERSION=3.0.1-rc2" \
```
wrkhenddher, line 85: Is it because we don't have a 3.0.1 yet?

.gitlab-ci.yml
```diff
@@ -82,6 +82,11 @@ Build and push image:
       docker build \
       --build-arg "NPM_REGISTRY=https://nexus.osdc.io/repository/npm-all/" \
```
wrkhenddher, line 83: Would it be too hard to get version and bake it as label?

https://github.com/NCI-GDC/gdcapi/blob/4dac2b9c63c746e17704c9633bf8d8cce0623c8c/docker.sh#L33-L37

Assuming it's simple to get version in a node app.

Dockerfile
```diff
@@ -1,10 +1,13 @@
-FROM node:20.11.0-alpine3.18 as dep
+ARG BASE_VERSION=3.0.1-rc2
```
wrkhenddher, line 1: same as previous.

Dockerfile
```diff
@@ -17,7 +20,7 @@ COPY ./package.json ./package-lock.json lerna.json ./
 COPY ./packages/core/package.json ./packages/core/
 COPY ./packages/sapien/package.json ./packages/sapien/
 COPY ./packages/portal-proto/package.json ./packages/portal-proto/
-RUN npm ci
+RUN npm ci --include=dev
```
wrkhenddher, line 23: why dev deps?

Isn't this image the one we used in prod?

https://docs.npmjs.com/cli/v10/commands/npm-install

Dockerfile
```diff
@@ -27,29 +30,32 @@ RUN lerna run --scope @nci-gdc/sapien build
...
+LABEL org.opencontainers.image.title=${NAME} \
+      org.opencontainers.image.description="${NAME} container image" \
+      org.opencontainers.image.source="https://github.com/NCI-GDC/${NAME}" \
+      org.opencontainers.image.vendor="NCI GDC"
+
 WORKDIR /app
 ENV NODE_ENV=production \
     PORT=3000
 
-RUN  addgroup --system --gid 1001 nextjs && adduser --system --uid 1001 nextjs
```
wrkhenddher, line 35: No nextjs user needed anymore?

Dockerfile
```diff
@@ -27,29 +30,32 @@ RUN lerna run --scope @nci-gdc/sapien build
...
+COPY --from=builder --chown=app:app /app/node_modules ./node_modules
+COPY --from=builder --chown=app:app /app/packages/portal-proto/public ./packages/portal-proto/public
+COPY --from=builder --chown=app:app /app/packages/portal-proto/package.json ./packages/portal-proto/package.json
+COPY --from=builder --chown=app:app /app/packages/portal-proto/.next ./packages/portal-proto/.next
+COPY --from=builder --chown=app:app /app/packages/portal-proto/node_modules ./packages/portal-proto/node_modules
+COPY --from=builder --chown=app:app /app/packages/portal-proto/next.config.js ./packages/portal-proto/next.config.js
 
 RUN mkdir -p ./packages/portal-proto/.next \
-  && chown nextjs:nextjs ./packages/portal-proto/.next
+  && chown app:app ./packages/portal-proto/.next
```
wrkhenddher, line 55: so UID:GID app:app exists at this point?

Without being added? Or is it part of the base image already?

wrkhenddher (approved):

Dockerfile
qqiao2024: npm ci is doing npm prepare which runs husky install and husky is in dev deps. 
There are many other dev deps (including lodash) used in lerna run --scope portal-proto build. 

Either some scripts are not well defined or some deps are not well defined. Not something we should solve here.

Dockerfile
qqiao2024: using predefined user `app`

Dockerfile
qqiao2024: yes

.gitlab-ci.yml
qqiao2024: right, this is very new.

.gitlab-ci.yml
qqiao2024: I thought we decided to remove all versions from labels because we are not rebuilding for tags, 
and the versions will just confuse.

Dockerfile
craigrbarnes: We can fix the dependencies and build flow in another ticket.

PR #981, opened 2024-02-19
PEAR-1536 - cDAVE - Adding multiple filters and selecting survival plot, box/qq plots cause performance slow down   

amlehman:
## Description
 Improve performance of cDave with large QQ plots by 
1) replacing Victory Charts with Apache eCharts which handles charts with large datasets more performantly 
2) reducing renders on the page with memoization and other techniques

## Checklist

- [x] Added proper unit tests
      flagged issues

Screenshot 2024-02-19 at 1 10 27 PM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,6 +15,7 @@
         "@reactour/tour": "^2.9.0",
         "classnames": "^2.3.1",
         "dom-to-svg": "^0.12.2",
+        "echarts": "^5.4.3",
         "file-saver": "^2.0.5",
         "html2canvas": "^1.4.1",
         "immutability-helper": "^3.1.1",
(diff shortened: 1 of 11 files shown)


wteouchicago:
Still testing, but is there a way to make the dots/points on the scatterplot unfilled circles rather than filled circles?

wteouchicago:
Looking good -- not spotting any major issues. Things do feel less sluggish.

For the unfilled circles thing, I tried something like the following, but no idea if it has bad side effects, e.g. if something changes the background of the viz to something non-white. 🤷‍♀️  If it's a pain, lemme know and I'll discuss rq with stakeholders.

```
 itemStyle: {
            opacity: 0.7,
            borderColor: color,
            color: "transparent",
            borderWidth: 2
          },
```
(huh. I tried "white" at first, but apparently "transparent" works too?)

Screenshot 2024-02-21 at 11 04 18 PM



### Issues
1. When the numbers on the y-axis are big, they overlap with the y-axis label. Can we tweak this to avoid the overlap since users may use these in publications? I think previously the fonts were smaller, which may help. Open to whatever works.
Screenshot 2024-02-21 at 10 35 04 PM

wteouchicago:
Looks good. verified that the label isn't overlapping with the values anymore and that the points on the plot are more like donuts than frisbees now. 👍 👍

paribartandhakal (approved):

PR #988, opened 2024-02-23
PEAR-1547: Cohort Builder styling issues

alilahrime:
## Description
[pear-1547](https://gdc-ctds.atlassian.net/browse/PEAR-1547)


wteouchicago:
@alilahrime , can you note which of the items on the ticket were attempted? As far as I can tell, items 2 and 4 are not changed?

I think for item 1, more padding may be needed. @cemile-ctds can comment on this one probably.

As for item 3: it's not looking like the screenshot in the ticket - the position and button style are wrong.

wteouchicago (changes requested):
see comment above

craigrbarnes (changes requested):
It is clear from the code that a subset of items are addressed.

Please add a screenshot for all the items listed in the PR so the reviewers know when all items are ready for review. 

If this PR is for a subset of items, please indicate which items need to be reviewed

alilahrime:
> see comment above

1. made the padding more apparent. didn't find a way to only apply it to specific facet cards, but now none of them look like they have zero padding at least.
2. there were 2 places that border is defined, changed the correct one this time
3. added centering and modified text/border color (just opted for primary here)
4. the card widths were defined based off different screen sizes so i made them always grid-col-4 instead. let me know if this is acceptable

wteouchicago:
For 1: There may be a little too much padding now, esp. for cards like Tobacco Smoking Onset Year. I'll let @cemile-ctds decide.

2 looks ok to me.

3: the background of the button doesn't look right. Also, once I add a custom card, the button and card are horizontally centered, and I don't think they are supposed to be.

Screenshot 2024-03-05 at 10 09 48 AM

4. The card widths are still inconsistent for me -- e.g. the cards in General vs the the ones in Treatment. This is happening at 1920 width (and other widths)

cemile-ctds:
1: It looks fine, but for consistent card spacing, it would be better to change the padding to `p-2` instead of `p-4`.
2: Everything seems good on my end.
3:The "Add a custom filter" button should have a white background color. Additionally, I think centering the "no custom filter added" affects the appearance of the custom filter tabs when they are added.
4: We prefer not to use a fixed grid, and I believe the issue is not related to different screen sizes because it occurs even when the user doesn't change the screen size. The width of the facet cards remains inconsistent across all tabs, regardless of whether the grid is fixed or not.
*******
5: The border-radius needs to be fixed for the cards like Cigarettes Per Day (Exposure):
Screenshot 2024-03-05 at 11 00 40
Screenshot 2024-03-05 at 11 10 28

cemile-ctds:
All issues listed above are fixed and they look good on my side.

craigrbarnes (approved):
Code changes are fine

wteouchicago (approved):

PR #992, opened 2024-02-27
PEAR-1561 : Cohort Builder non-days range values being converted to years

alilahrime:
## Description
[pear-1561](https://gdc-ctds.atlassian.net/browse/PEAR-1561)


wteouchicago:
Seems to work for me. Could you add some unit tests since we have tests for `adjustYearsToDays` but I didn't see any for `adjustDaysToYears`?

alilahrime:
> adjustYearsToDays

truee. added some tests for `adjustDaysToYears`

packages/portal-proto/src/features/facets/utils.unit.test.ts
```diff
@@ -501,3 +502,17 @@ describe("test years to days conversion", () => {
...
+    expect(adjustDaysToYears(182, "years")).toEqual(0);
+    expect(adjustDaysToYears(183, "years")).toEqual(1);
+    expect(adjustDaysToYears(364, "days")).toEqual(364);
+    expect(adjustDaysToYears(730, "years")).toEqual(2);
+    expect(adjustDaysToYears(36500, "years")).toEqual(100);
+    expect(adjustDaysToYears(32873, "years")).toEqual(90);
+    expect(adjustDaysToYears(-25568, "years")).toEqual(-70);
+    expect(adjustYearsToDays(-32873, "days")).toEqual(-32873);
+  });
+});
```
paribartandhakal: should it not be 365 days? Is the function correct?

wteouchicago: yeah, from the name, i thought this converts from days to years as appropriate. if so, then most of the "years" should be "days" (and some of the "days" should be "years"? 🤷‍♀️

alilahrime: its a bit confusing but the conversion should only happen if the enum/unit is `years` (plural). otherwise return the value itself when its `days`, `year` (singular), any other enum that the shared facet component consumes

packages/portal-proto/src/features/facets/utils.unit.test.ts
```diff
@@ -501,3 +502,17 @@ describe("test years to days conversion", () => {
...
+  test("days to years", () => {
+    expect(adjustDaysToYears(365, "years")).toEqual(1);
+    expect(adjustDaysToYears(182, "years")).toEqual(0);
+    expect(adjustDaysToYears(183, "years")).toEqual(1);
+    expect(adjustDaysToYears(364, "days")).toEqual(364);
+    expect(adjustDaysToYears(730, "years")).toEqual(2);
+    expect(adjustDaysToYears(36500, "years")).toEqual(100);
+    expect(adjustDaysToYears(32873, "years")).toEqual(90);
+    expect(adjustDaysToYears(-25568, "years")).toEqual(-70);
+    expect(adjustYearsToDays(-32873, "days")).toEqual(-32873);
```
wteouchicago: was this supposed to be `adjustDaysToYears`?

alilahrime: yes! my mistake, good catch

packages/portal-proto/src/features/facets/utils.unit.test.ts
```diff
@@ -501,3 +502,17 @@ describe("test years to days conversion", () => {
     expect(adjustYearsToDays(90, "years")).toEqual(32873);
   });
 });
+
+describe("test days to years conversion", () => {
+  test("days to years", () => {
```
craigrbarnes, line 507: Can you add more test cases to cover:
range, age, year, percent?
Just to complete the coverage of the possible values of units. One each would be fine. If `adjustDaysToYears` adds  a case for unit=="percent" the test pick it up.

craigrbarnes (changes requested):
See my comment on utils.unit.test.ts

craigrbarnes (changes requested):
Could you also rename the functions: `adjustYearsToDays` and `adjustDaysToYears` to 
`adjustYearsToDaysIfUnitsAreYears` and `adjustDaysToYearsIfUnitsAreYears`. I know it's wordy but it will be more apparent what is happening when these are called.

packages/portal-proto/src/features/facets/utils.unit.test.ts
alilahrime: sounds good. added a few for each of those units

craigrbarnes (approved):
The code was changed to only adjust years <-> days if the facet's unit is "years"

PR #1018, opened 2024-03-18
PEAR-1844 - When clicking a link to a case summary modal, the background page scrolls to the bottom

amlehman:
## Checklist

- [ ] Added proper unit tests
      flagged issues


wteouchicago:
works for me 👍 

qn: I don't think we have a link like this anywhere yet, but what happens if the case summary page is opened in a modal and the `bioId` is provided -- will it still scroll to the right spot?

amlehman:
Hi @wteouchicago, good question. I looked into this a bit and we would need to do some more work to have the scrolling work in the modal. I've changed my fix a bit to be more amenable to adding that functionality though.

craigrbarnes (approved):

PR #1042, opened 2024-03-28
PEAR-661: hide range control if there is no data

craigrbarnes:
## Description
Hides the custom range controls when the facet does not have data. There is a change from the ticket:
If the custom range value does not result in data for that facet, the custom range is shown; otherwise, the user can only clear the filters from the query expression.

Note: the Age facets have the Days/Years Segment rendered in the parent component. A handler is now passed to `RangeInputWithPrefixedRanges` to hide it if there are no values. This should be refactored later.

This also includes undefined value fixes with custom facets.

## Checklist

- [ ] Added proper unit tests
      flagged issues

### Example of Year of Diagnosis with no data
image

### Example of Year of Diagnosis with values that result in no data:
image

### Age of Diagnosis with no data:
image

### DIFF
--- a/packages/portal-proto/src/features/facets/NumericRangeFacet.tsx
+++ b/packages/portal-proto/src/features/facets/NumericRangeFacet.tsx
@@ -671,6 +671,7 @@ const DaysOrYears: React.FC<NumericFacetData> = ({
 }: NumericFacetData) => {
   const [units, setUnits] = useState("years");
   // no data if true means the Day/Year SegmentedControl should not be rendered.
+  // TODO: this is not ideal and perhaps should be refactored
   const [noData, setNoData] = useState(false);
   // set up a fixed range -90 to 90 years over 19 buckets
   const rangeMinimum = -32873;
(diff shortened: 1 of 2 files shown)


wteouchicago:
functionality looks good. 👍 
any chance we can add a bit of padding to the bottom (below the "No data for this field" text)? Or would https://github.com/NCI-GDC/gdc-frontend-framework/pull/988 take care of it?

craigrbarnes:
> functionality looks good. 👍 any chance we can add a bit of padding to the bottom (below the "No data for this field" text)? Or would #988 take care of it?

I will fix it in this one.

packages/portal-proto/src/features/facets/NumericRangeFacet.tsx
```diff
@@ -546,6 +548,21 @@ const RangeInputWithPrefixedRanges: React.FC<
     setIsGroupExpanded(!isGroupExpanded);
   };
 
+  // informs the parent component if there is data or no data
+  // only used by the DaysOrYears component
+  useEffect(() => {
```
paribartandhakal: I think it's better to use `useDeepCompareEffect` given that `filterValues` is an object. Also, I feel like we are updating other `useEffect` -> `useDeepCompareEffect` as we work on things gradually. I am not sure if it has any performance impact.

paribartandhakal: Also a reminder to add padding as Wendy suggested. (commenting as i didn't see it in the newer commits)

craigrbarnes: Agreed on both. I will do that

craigrbarnes: Both suggestions are completed.

paribartandhakal (approved):

PR #1046, opened 2024-03-29
prototype using GeneSetModal in OncoMatrix

siosonel:
## Description

Addresses https://gdc-ctds.atlassian.net/browse/FEAT-766.

Test by opening OncoMatrix, wait to laod, then click Genes -> Edit Group -> Load Gene Sets -> (enter list/save/submit).

@craigrbarnes I'm not able to click `Cancel` or `X` icon to close the gene set modal, please advise.

## Checklist

- [ ] Added proper unit tests
      flagged issues


siosonel:
@craigrbarnes I've started to use GeneSetModal in `OncoMatrixWrapper.tsx`, following the pattern in `FilterPanel.tsx`.  However, I'm seeing errors that may be due to improper use or the GeneSetModal. Please note that testing this branch requires having a local PP dev environment, so the errors reported here mat not be easily replicated with the standard GFF dev setup. 

Screenshot 2024-03-29 at 10 42 15 AM

craigrbarnes:
> @craigrbarnes I've started to use GeneSetModal in `OncoMatrixWrapper.tsx`, following the pattern in `FilterPanel.tsx`. However, I'm seeing errors that may be due to improper use or the GeneSetModal. Please note that testing this branch requires having a local PP dev environment, so the errors reported here mat not be easily replicated with the standard GFF dev setup.
> 
> Screenshot 2024-03-29 at 10 42 15 AM

I just saw this. I will investigate now

craigrbarnes:
> `ncoMatrixWrapper.tsx`,

I see what the issue is: The MF app has its own redux store, as we have implemented it as an app with its own local filters. You can see it defined here:  https://github.com/NCI-GDC/gdc-frontend-framework/blob/6667b824b94d2a08e0b9da49896cee2dffcc2604/packages/portal-proto/src/features/genomic/appApi.ts

This means that this hook can only be used by the Mutation Frequency app and not anywhere else as it will not have access to MF app's redux store. 

The developer's guide describes how apps are structured and implemented.

siosonel:
@craigrbarnes I was able to get back to this task only this week. After looking at your code links and usage patterns, it's now clear to me that the `GeneSetModal` is meant to update a general or app-specific (cohort) filter via hooks. Whereas, the PP use case is to simply get a gene list to use a rows in the OncoMatrix, it's not meant to filter the cohort. Would it be possible for the `GeneSetModal` to have a calback option that receives a gene list/array as argument, instead of triggering hooks? Or maybe there is a better solution or another modal that can do this? Basically, we'd like to reuse a modal that a GDC user is familiar with, that would allow recovering their saved gene set, not necessarily a filter although we can parse the gene list from that as needed.

craigrbarnes:
See my comment here https://gdc-ctds.atlassian.net/browse/FEAT-766?focusedCommentId=62740
That call should be what you need. For the GeneSetModal, you could pass a custom `updateFilters` function, which you could use to get the selected gene set ids. If that is not what you need, let us know.

siosonel:
@craigrbarnes ~~would it be okay to rebase this long-lived branch on `develop` so that the PR discussion/history is kept, or should I open another PR with a new branch that's already based off develop *and just refer to this PR)?~~ will just reuse this branch/PR

craigrbarnes (approved):

craigrbarnes:
I approved. However, the GitLab pipeline status needs to be clear. I tried an empty commit, but it did not work. I will work with devops to see what the issue is.

PR #1048, opened 2024-04-01
Qa 2177 automation ids summary pages

bpierce01:
## Description
- Created automation IDs for following summary pages: project, case, file, gene, mutation, annotation

### DIFF
--- a/packages/portal-proto/src/components/DownloadButtons/DownloadFile.tsx
+++ b/packages/portal-proto/src/components/DownloadButtons/DownloadFile.tsx
@@ -17,6 +17,7 @@ interface DownloadFileProps {
   setfileToDownload?: React.Dispatch<React.SetStateAction<GdcFile>>;
   customStyle?: string;
   showLoading?: boolean;
+  customDataTestID?: string;
 }
 
 export const DownloadFile: React.FC<DownloadFileProps> = ({
@@ -26,6 +27,7 @@ export const DownloadFile: React.FC<DownloadFileProps> = ({
   setfileToDownload,
   customStyle,
   showLoading = true,
+  customDataTestID,
 }: DownloadFileProps) => {
   const dispatch = useCoreDispatch();
   const [fetchUserDetails] = useLazyFetchUserDetailsQuery();
@@ -64,6 +66,7 @@ export const DownloadFile: React.FC<DownloadFileProps> = ({
   if (file.access === "open") {
     return (
       <DownloadButton
+        data-testid={customDataTestID}
         extraParams={{
           ids: file.file_id,
           annotations: true,
@@ -84,6 +87,7 @@ export const DownloadFile: React.FC<DownloadFileProps> = ({
 
   return (
     <DownloadButton
+      data-testid={customDataTestID}
       customStyle={customStyle || customStyleFile}
       inactiveText={inactiveText}
       activeText={activeText}
--- a/packages/portal-proto/src/components/Summary/CategoryTableSummary.tsx
+++ b/packages/portal-proto/src/components/Summary/CategoryTableSummary.tsx
@@ -9,6 +9,7 @@ interface CategoryTableSummaryProps<TData> {
   data: TData[];
   columns: ColumnDef<TData, any>[];
   tooltip?: string;
+  customDataTestID?: string;
 }
 
 function CategoryTableSummary<TData>({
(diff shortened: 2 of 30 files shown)


packages/portal-proto/src/components/DownloadButtons/DownloadFile.tsx
```diff
@@ -59,6 +61,7 @@ export const DownloadFile: React.FC<DownloadFileProps> = ({
   if (file.access === "open") {
     return (
       <DownloadButton
+        data-testid={customDataTestID}
```
paribartandhakal, line 64: Is this part working? Are you able to see the data-testid?

packages/portal-proto/src/components/DownloadButtons/DownloadFile.tsx
```diff
@@ -79,6 +82,7 @@ export const DownloadFile: React.FC<DownloadFileProps> = ({
 
   return (
     <DownloadButton
+      data-testid={customDataTestID}
```
paribartandhakal, line 85: Is this part working? Are you able to see the data-testid?

packages/portal-proto/src/components/DownloadButtons/DownloadFile.tsx
bpierce01: Yes it appears for open/controlled access file summaries and for the little download file buttons

packages/portal-proto/src/components/DownloadButtons/DownloadFile.tsx
bpierce01: yes, see above

paribartandhakal (approved):

packages/portal-proto/src/components/DownloadButtons/DownloadFile.tsx
uc-pperumala: @bpierce01 can you give me an example on how this transforms to for the element?

uc-pperumala: I feel like we would have to set a prefix, such as `button-${customDataTestID}`. Please correct me if I am wrong.

packages/portal-proto/src/components/Summary/CategoryTableSummary.tsx
```diff
@@ -29,12 +31,20 @@ function CategoryTableSummary<TData>({
...
+                size={16}
+                className="text-accent"
+              />
             </ActionIcon>
           </Tooltip>
         )}
       </div>
-      <VerticalTable data={data} columns={columns} />
+      <VerticalTable
+        customDataTestID={customDataTestID}
```
uc-pperumala, line 44: Same here, will need to set up a prefix such as `text-${customDataTestID}` or something similar?

packages/portal-proto/src/components/Table/VerticalTable.tsx
```diff
@@ -98,6 +99,7 @@ function VerticalTable<TData>({
   setExpanded,
   getRowId = getDefaultRowId,
   baseZIndex = 0,
+  customDataTestID,
```
uc-pperumala, line 102: Just in general, how is the value being set for this variable.

uc-pperumala (approved):
Just left a few clarifying questions, but otherwise, looks good to me!

packages/portal-proto/src/components/Table/VerticalTable.tsx
bpierce01: Pari could explain more eloquently, but for example VerticalTable is an object that is used throughout the framework. It is a custom object, and we cannot pass a data-testid directly into it. So, we create the variable 'customDataTestID' and then in the object we write data-testid={customDataTestID} at the appropriate location. So then, when we see VerticalTable being used in the framework, we can pass in a customDataTestID at that point and then that gets presented as a data-testid.

packages/portal-proto/src/components/Summary/CategoryTableSummary.tsx
bpierce01: This one was more complex where I needed to pass in a 'customDataTestID' through two layers before it shows up as the data-testid

craigrbarnes (approved):
Changes are fine. Tested tables and other components. updated in this PR, and see no issues

PR #1050, opened 2024-04-01
PEAR-1866 -Browse Annotations - Filters persist when user leaves page and comes back

amlehman:
## Checklist

- [x] Added proper unit tests
      flagged issues

### DIFF
--- a/packages/portal-proto/src/features/annotations/AnnotationBrowser.tsx
+++ b/packages/portal-proto/src/features/annotations/AnnotationBrowser.tsx
@@ -1,7 +1,14 @@
+import React, { useEffect } from "react";
 import AnnotationFilterPanel from "./AnnotationFilterPanel";
 import AnnotationTable from "./AnnotationTable";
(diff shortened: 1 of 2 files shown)


wteouchicago:
looks good.

qn: is it possible not to clear the filters on refresh, but to clear them on leaving the page and returning? Or are they both the same "mechanism"? for example, the projects app doesn't clear filters on refresh, only on closing the app and returning, but I get that maybe it's different cos it's an app and this is a page.

amlehman:
Hi @wteouchicago, I could set the annotation app to persist the filters through refresh. Though the way I'm clearing the filters at the moment, it would have the same issue noted in this ticket: https://gdc-ctds.atlassian.net/browse/PEAR-1763. There may be a different approach to clearing filters that will solve that issue, I haven't looked into it yet.

wteouchicago:
Thanks @amlehman -- let's leave this alone for now. We can figure out some consistent behaviour across the portal later. 

All good here. 👍

craigrbarnes (approved):

PR #1053, opened 2024-04-02
update the proteinpaint-client version to 2.50.0

siosonel:
## Description

Also,  handle empty userDetails.data in sequence read app

Features:
- include auth test status in server healthcheck
- Hide synonymous mutations by default for GDC oncoMatrix
- Improve the matrix sorting options to easily toggle sorting by cnv and/or consequence

Fixes:
- fix the error from genomic alterations rendering when there are no mutations or CNV data
- In GDC query, do not supply empty "case_filters{content[]}" that will slow down API. lollipop and oncomatrix are now faster when there's no cohort
- at GDC bam slicing UI, the table listing available cases and bam files can be filtered by assay types
- Add to GDC oncoMatrix mutation/cnv buttons all available mutation/cnv classes in GDC instead of all available mutation/cnv classes in the current matrix

## Checklist

- [ ] Added proper unit tests
      flagged issues

### DIFF
--- a/packages/portal-proto/src/features/proteinpaint/SequenceReadWrapper.tsx
+++ b/packages/portal-proto/src/features/proteinpaint/SequenceReadWrapper.tsx
@@ -1,4 +1,5 @@
-import { useEffect, useRef, useState, FC } from "react";
+import { useRef, useState, FC } from "react";
+import { useDeepCompareEffect } from "use-deep-compare";
 import { runproteinpaint } from "@sjcrh/proteinpaint-client";
 import {
   useCoreSelector,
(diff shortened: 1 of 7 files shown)


packages/portal-proto/src/features/proteinpaint/SequenceReadWrapper.tsx
```diff
@@ -34,7 +34,7 @@ export const SequenceReadWrapper: FC<PpProps> = (props: PpProps) => {
   useEffect(
```
craigrbarnes: we have been using `useDeepCompareEffect`, especially when the dependencies are objects. I suggest using it here. The library's homepage is: https://github.com/sandiiarov/use-deep-compare#readme

craigrbarnes (changes requested):
See my comment about `useDeepCompareEffect`.

packages/portal-proto/src/features/proteinpaint/SequenceReadWrapper.tsx
siosonel: Ok, thanks, I've replaced with `useDeepCompareEffect()`.

craigrbarnes: Thank you. I will merge it now

craigrbarnes (approved):

PR #1056, opened 2024-04-03
PEAR-1740 - Cohort Bar Case Table - Search can sometimes not happen

amlehman:
## Description
This one was a race condition. 

## Checklist

- [x] Added proper unit tests
      flagged issues

### DIFF
--- a/packages/core/src/features/cases/casesSlice.ts
+++ b/packages/core/src/features/cases/casesSlice.ts
@@ -191,6 +191,7 @@ export const fetchAllCases = createAsyncThunk<
 export interface CasesState {
   readonly allCasesData: CoreDataSelectorResponse<CaseSliceResponseData[]>;
   readonly totalSelectedCases?: number;
+  readonly requestId?: string;
 }
 
 const initialState: CasesState = {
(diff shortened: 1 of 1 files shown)


wteouchicago:
works for me! 👍

packages/core/src/features/cases/casesSlice.ts
```diff
@@ -206,6 +207,8 @@ const slice = createSlice({
   extraReducers: (builder) => {
     builder
       .addCase(fetchAllCases.fulfilled, (state, action) => {
+        if (state.requestId != action.meta.requestId) return state;
```
paribartandhakal, line 210: Just wondering how does this help with the solution?

amlehman: The issue is basically that sometimes the search results will return out of order (particularly if you kick off a large search and then do a search with a small amount of results that returns quickly.) This prevents the state from getting updated by the earlier request. We're using this pattern on some other slices for this same reason: https://github.com/NCI-GDC/gdc-frontend-framework/blob/develop/packages/core/src/features/annotations/annotationsSlice.ts#L47

paribartandhakal (approved):

craigrbarnes (approved):

PR #1063, opened 2024-04-05
PEAR-869 removes no description and empty tooltips

wteouchicago:
## Description
- disables/removes tooltips rather than show "no description" or similar, and when it was showing empty tooltips

## Checklist

      flagged issues


packages/portal-proto/src/features/cDave/Controls.tsx
```diff
@@ -193,6 +193,7 @@ const FieldControl: React.FC<FieldControlProps> = ({
           </Highlight>
         </>
       ) : (
+        // TODO don't show No desc available when tooltip is working correctly again
```
wteouchicago: I left this one alone for now because the tooltip is acting oddly (PEAR-1878).

wteouchicago:
Bah - failing tests

packages/portal-proto/src/features/annotations/filters.json
```diff
@@ -3,7 +3,7 @@
     "title": "Project",
     "field": "project",
     "full": "project.project_id",
-    "description": "No Description",
+    "description": null,
```
wteouchicago: fwiw, I don't think setting to `null` is great, but wanted to get around errors like:

> src/features/annotations/AnnotationFilterPanel.tsx(54,13): error TS2352: Conversion of type '{ title: string; field: string; full: string; type: string; doc_type: string; facet_type: string; }' to type 'FacetDefinition' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.

I would rather set description to optional but wasn't sure if there was a reason it's required.

packages/portal-proto/src/features/genomic/filters.json
```diff
@@ -14,14 +14,13 @@
   {
     "facet_filter": "genes.biotype",
     "name": "Biotype",
-    "description": "No description",
+    "description": null,
```
wteouchicago: this was to get around: 
> src/features/genomic/FilterPanel.tsx(93,30): error TS2339: Property 'description' does not exist on type '{ facet_filter: string; name: string; type: string; docType: string; }'.

please let me know if there is a better way.

packages/portal-proto/src/features/annotations/filters.json
amlehman: The FacetDefinition type is more describing what we get back from the _mapping endpoint. I'm not sure if they always have descriptions there? If not, it makes sense to have description as optional in the type. If we leave description as required, we can have the props for CreateFacetCard look for partial matches to the type like this:
```
index cde91864..f8a6443b 100644
--- a/packages/portal-proto/src/features/annotations/AnnotationFilterPanel.tsx
+++ b/packages/portal-proto/src/features/annotations/AnnotationFilterPanel.tsx
@@ -51,7 +51,7 @@ export const AnnotationFacetPanel = (): JSX.Element => {
         {FilterFacets.map((x) => {
           const facetName = x.title || fieldNameToTitle(x.full);
           return createFacetCard(
-            x as FacetDefinition,
+            x as Partial<FacetDefinition>,
             "Annotations",
             facetHooks,
             "annotation-browser",
diff --git a/packages/portal-proto/src/features/facets/CreateFacetCard.tsx b/packages/portal-proto/src/features/facets/CreateFacetCard.tsx
index ccde12cb..cb37e252 100644
--- a/packages/portal-proto/src/features/facets/CreateFacetCard.tsx
+++ b/packages/portal-proto/src/features/facets/CreateFacetCard.tsx
@@ -25,7 +25,7 @@ import { FacetRequiredHooks } from "@/features/facets/types";
  * @param width - override width of facet
  */
 export const createFacetCard = (
-  facet: FacetDefinition,
+  facet: Partial<FacetDefinition>,
   valueLabel: string,
   dataFunctions: FacetRequiredHooks,
   idPrefix: string,
amylehman@Amys-MBP gdc-frontend-framework % 
```

wteouchicago: I find it hard to believe that the _mapping endpoint will always return a description. Yet it's odd that `description` has been set to required and tests were passing. I'll investigate more with BE and decide which way to go based on findings. 

🙇‍♀️

wteouchicago: Ended up taking the `<Partial>` path. Seems like mappings will return an empty string for `description` whenever we don't actually have a description for the property.

packages/portal-proto/src/features/genomic/filters.json
wteouchicago: per pari's suggestion, I ended up removing `description` from the cards in Mutation Frequency since we don't have any descriptions. Figured we can add it back in if we ever have descriptions.

amlehman (approved):

craigrbarnes (approved):

PR #1066, opened 2024-04-05
(PEAR-1301): update icons, update header styles to use icons, update search to use icons

ocshawn:
[PEAR-1301](https://gdc-ctds.atlassian.net/browse/PEAR-1301)

## Description
- update icons
- update header styles to use icons
- update search to use icons

## Checklist

- [x] Added proper unit tests
      flagged issues


cemile-ctds (approved):
The style changes and icon placement look good on my end.

wteouchicago:
### Some issues in Quick Search:

1. The first row of text (where it's bolded) seems to be wrapping when it doesn't need to be. Any reason for it?
Screenshot 2024-04-09 at 5 16 58 PM

2. For the category, can we display "Mutation" instead of "Ssm"? 
Screenshot 2024-04-09 at 5 22 36 PM

3. More of a question than an issue. Did we intentionally remove the bolding of text that match? I'm fine either way and I see we don't bold in the Cohort Builder search.  @cemile-ctds ?

Screenshot 2024-04-09 at 5 24 57 PM

ocshawn:
For 3. i removed it to match @cemile-ctds what would you like?

everything else should be fixed @wteouchicago

cemile-ctds:
The cohort builder search doesn't change the font style of the highlighted, matching text. Let's keep it the same as that.

wteouchicago:
looks great, thanks! 👍

packages/portal-proto/src/components/QuickSearch/QuickSearch.tsx
```diff
@@ -179,7 +184,8 @@ export const QuickSearch = (): JSX.Element => {
       aria-label="Quick Search Input"
       classNames={{
         input: "focus:border-2 focus:border-primary text-sm",
-        dropdown: "bg-base-lightest border-r-10 border-1 border-base",
+        dropdown:
+          "bg-base-max rounded-t-none rounded-b border-0 drop-shadow-md mt-[-8px]",
```
paribartandhakal: We could probably do `-mt-2`.

packages/portal-proto/src/components/QuickSearch/QuickSearch.tsx
```diff
@@ -96,56 +106,51 @@ export const QuickSearch = (): JSX.Element => {
...
-                        : "bg-primary-darker text-primary-contrast-darker"
-                    }`,
-                  }}
-                  className="cursor-pointer"
-                >
-                  {badgeText}
-                </Badge>
+            <div
+              className={`py-2 flex gap-2 ${
+                last ? "" : "border-b border-gdc-grey-light"
```
paribartandhakal, line 129: Nitpick: Do you think whether we can leverage `last` pseudo class here: https://tailwindcss.com/docs/hover-focus-and-other-states#first-last-odd-and-even?

packages/portal-proto/src/components/QuickSearch/entityShortNameMapping.ts
```diff
@@ -12,18 +19,21 @@ export type QuickSearchEntities =
...
-export const entityShortNameMapping: Record<QuickSearchEntities, string> = {
-  case: "CA",
-  file: "FL",
-  project: "PR",
-  annotation: "AN",
-  gene_centric: "GN",
-  ssm_centric: "MU",
+export const entityIconMapping: Record<
+  QuickSearchEntities,
+  { icon: any; category?: string }
```
paribartandhakal, line 24: We can probably use this type `import { IconType } from "react-icons";`.

packages/portal-proto/src/components/Summary/SummaryHeader.tsx
```diff
@@ -3,20 +3,20 @@ import { Divider } from "@mantine/core";
 import { ReactNode } from "react";
 
 export interface SummaryHeaderProps {
-  iconText: string;
+  Icon?: any;
```
paribartandhakal: Here too we could use the icon type.

packages/portal-proto/src/components/QuickSearch/QuickSearch.tsx
ocshawn: the problem is its a sub element, do you know if there is a way to handle for that?

craigrbarnes (approved):

PR #1076, opened 2024-04-10
PEAR-1843 - Add custom view names

amlehman:
\+ PEAR-1890 - Disable DataDog RUM in non-production modes
\+ PEAR-1889 - Configure DataDog RUM based on version

## Description
To test, you need to run the app in production mode. Where you would normally run npm run dev in the console, instead:
```
export NEXT_PUBLIC_GDC_APP_API_AUTH=https://portal.gdc.cancer.gov/auth/api/v0
export NEXT_PUBLIC_APP_VERSION=2.12.0
export NEXT_PUBLIC_APP_HASH=ba9feb5f 
npm run build
npm run start 
```
The version and hash are added during our build process normally but to see them show up testing locally, we need to set them manually. 

You can see the data that is sent to Datadog in requests to https://rum.browser-intake-datadoghq.com. They fire periodically and batch together actions so you may need to wait a few seconds to see the actions you are looking for:
Screenshot 2024-04-10 at 3 23 26 PM

To test that we aren't sending telemetry to Datadog in dev mode and that there aren't any errors, use:
```
npm run dev
```

## Checklist

- [x] Added proper unit tests
      flagged issues


packages/portal-proto/src/pages/_app.tsx
```diff
@@ -34,22 +34,23 @@ import { Notifications } from "@mantine/notifications";
...
-  // version: '1.0.0',
-  sessionSampleRate: 100,
-  sessionReplaySampleRate: 0,
-  trackUserInteractions: true,
-  trackFrustrations: true,
-  trackResources: true,
-  trackLongTasks: true,
-  defaultPrivacyLevel: "mask",
-});
+if (process.env.NODE_ENV === "production") {
```
philtom-ctds: This works when SSR executes this.  I'm not sure if this would get executed in the browser, but if it did, we'd need to use a `NEXT_PUBLIC_*` configuration. maybe something like
```
NEXT_PUBLIC_DD_ENABLED = process.env.NODE_ENV === "production"
```

packages/portal-proto/src/pages/404.tsx
```diff
@@ -1,10 +1,15 @@
...
 import Head from "next/head";
+import { datadogRum } from "@datadog/browser-rum";
 import { UserFlowVariedPages } from "@/features/layout/UserFlowVariedPages";
 import { headerElements } from "@/features/user-flow/workflow/navigation-utils";
 import Link from "next/link";
 
 const Custom404Page: NextPage = () => {
+  datadogRum.startView({
+    name: "Not Found",
+  });
```
philtom-ctds, line 11: Do you know how this behaves if we do not initialize datadog?  I'm hoping it silently does nothing as opposed to throwing an error.

packages/portal-proto/src/pages/annotations/index.tsx
```diff
@@ -10,6 +11,10 @@ const AnnotationsPage: NextPage = () => {
     selectGdcAppById(AnnotationBrowserAppId),
   ) as React.ElementType;
 
+  datadogRum.startView({
+    name: "Annotation",
```
philtom-ctds: nitpick:  could we make this plural, `Annotation Browser`?

packages/portal-proto/src/pages/404.tsx
amlehman: Yeah, I've tested it without initializing datadog and it is just ignored.

packages/portal-proto/src/pages/annotations/index.tsx
amlehman: Done 👍

philtom-ctds (approved):

PR #1078, opened 2024-04-11
PEAR-1839 - Add annotations table to case summary page, adjust links

amlehman:
## Checklist

- [x] Added proper unit tests
      flagged issues

Screenshot 2024-04-11 at 4 00 55 PM


wteouchicago:
Looks great! 👍 

Minor: can you remove the "Total of x annotations" text that is above the table? Admittedly we aren't very consistent for when we have it and when we don't -- I'll prob. have a follow-up ticket to make this consistent across the portal.

packages/core/src/features/gdcapi/gdcapi.ts
```diff
@@ -733,6 +733,18 @@ export const endpointSlice = coreCreateApi({
       }),
       transformResponse: (response) => response.data.hits,
     }),
+    getAnnotations: builder.query<
```
amlehman, line 736: Using the existing annotations hook was giving me an infinite loop when it got called in both the CaseView and AnnotationTable components :/

paribartandhakal (approved):
Looks good (need to remove table title as per above comment)

PR #1087, opened 2024-04-16
PEAR-1853 - Add annotations table to project summary page, adjust links

amlehman:
## Description
\+ PEAR-1855 - Add annotations table to file summary page, adjust links

Case ID is missing on the file tables right now due to: https://gdc-ctds.atlassian.net/browse/DEV-2653

## Checklist

- [x] Added proper unit tests
      flagged issues

Screenshot 2024-04-16 at 11 30 13 AM
Screenshot 2024-04-16 at 11 31 01 AM


wteouchicago:
Looking good!

Issues:
1. Minor: the project's annotations table should hide Case UUID column by default
2. Not sure if this is expected or if there are best practices in this area. If the project page is loaded in a modal and I click on the numAnnotations link in the header, "#annotations" is added in the URL, e.g. http://localhost:3000/analysis_page?app=Projects#annotations. I see the same is happening for anchor links in case summary page too. If it's not expected, I can file a separate ticket.
3. I can file a follow-up instead. When the project page is loaded with #annotations, can it automatically scroll down to the Annotations section (just like https://gdc-ctds.atlassian.net/browse/PEAR-1898). I guess we need to do the same for the case summary page. Lemme know if a separate jira is preferred.

wteouchicago:
4. Here's something odd. When I load http://localhost:3000/cases/7248cd60-be22-44bc-bc58-f644db0940a2#annotations , it scrolls to the Annotations section, then the mutations table finally loads and the Annotations section is shoved down. Not sure if this is worth fixing since not all cases will have annotations, etc. -- your call. 

Otherwise, looks good! 👍 

Edit: we won't be addressing 4.

craigrbarnes (approved):
Changes look fine

PR #1102, opened 2024-04-29
PEAR-1860 - Standardize data fetching slices (part 2)

amlehman:
## Description
- Adds the data release version to the cache key so slices will refetch when a data release happens
- Consolidates some redundant entity endpoints and types
- Removes some outdated set + cohort code

## Checklist

- [x] Added proper unit tests
      flagged issues


packages/core/src/features/quickSearch/quickSearch.ts
```diff
@@ -1,97 +1,40 @@
...
-      `${GDC_APP_API_AUTH}/quick_search?query=${searchString}&size=5`,
-    );
-    if (response.ok) {
-      return await response.json();
-    }
-
-    throw Error(await response.text());
+export const fetchQuickSearch = async (searchString: string) => {
+  const response = await fetch(
+    `${GDC_APP_API_AUTH}/quick_search?query=${searchString}&size=5`,
```
paribartandhakal: should we make a new const variable as `MAX_SEARCH_ITEMS` or something?

packages/core/src/serializeQueryArgs.ts
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  const dataRelease = Cookies.get("gdc_api_version");
+
+  return dataRelease
+    ? `${dataRelease}:${endpointName}(${serialized})`
+    : `${endpointName}(${serialized})`;
+};
+
+export default serializeQueryArgsWithDataRelease;
```
paribartandhakal, line 49: Is the purpose of the `serializeQueryArgsWithDataRelease` function to prevent` RTK Query` from using cached responses for API calls with different data release versions?

packages/portal-proto/src/components/ImageViewer/MultipleImageViewer.tsx
```diff
@@ -70,19 +67,13 @@ export const MultipleImageViewer = ({
...
     case_id,
     caseFilters: filters,
   });
 
-  useEffect(() => {
-    return () => {
-      dispatch(resetEdgesState());
-    };
-  }, [dispatch]);
-
```
paribartandhakal: This code was added to ensure that when the user navigates away from the image viewer tab (unmounts) and later returns to it, the tab will not display more than 10 slides upon remounting IIRC.  Will it still support this?

packages/core/src/features/quickSearch/quickSearch.ts
craigrbarnes: I agree. It would be good to have this as a constant, even if we never change it, is describe what that value is.

craigrbarnes (commented):
There is an issue with the image loader:
The Show More button does not load the next 10 images.

1. Go to repository
2. Select View Images
3. Click Show More:
4. The next 10 should load, but does not ( 20 of 430, 30 or 430...)

wteouchicago:
Tested these general areas, lemme know if I've missed something.

Annotations Browser, case summary page, cases table, cDAVE, Repo/Cart, Gene and Mutation summary pages, Image viewer, projects app + summary page, quick search, sets/manage sets, cohort builder.

### Issues
1. Manage Sets -- sometimes the numItems side panel shows the list of entities from a different set. See attachment. Steps:
- make 2 sets, one with at least 50 items (possibly the more the merrier, aka bigSet)
- click the # items button for bigSet. Scroll a little in the panel to let more items load (this seems to be how I often get the issue to show up).
- close the panel, click # items for the other set
- note that the items from bigSet are still shown

Not sure if this is due to this PR or slow local machine or...

https://github.com/NCI-GDC/gdc-frontend-framework/assets/73256434/0e24ee14-1a9e-48ae-a9bd-f20506e67dcc

packages/core/src/serializeQueryArgs.ts
amlehman: Yep, you've got it. I can add a comment in this file to clarify.

packages/portal-proto/src/components/ImageViewer/MultipleImageViewer.tsx
amlehman: I'm not sure that would be an issue with how the rtk query caching works but I'll check.

wteouchicago:
Issue 1 is fixed! 👍  all good here. 👍

craigrbarnes (approved):
ImageViewer now working as expected. Code changes are fine.

PR #1109, opened 2024-05-06
PEAR-1911 - Convert cohorts created via analyses into static cohorts

amlehman:
## Description
\+ PEAR-1885 - Move set creation to when user saves cohort

## Checklist

- [x] Added proper unit tests
      flagged issues

### DIFF
--- a/packages/portal-proto/src/components/CohortCreationButton.tsx
+++ b/packages/portal-proto/src/components/CohortCreationButton.tsx
@@ -1,5 +1,5 @@
 import React, { ReactNode, useState } from "react";
-import { Loader, Tooltip } from "@mantine/core";
+import { Tooltip } from "@mantine/core";
 import { FaPlus as PlusIcon } from "react-icons/fa";
 import tw from "tailwind-styled-components";
 import { FilterSet } from "@gff/core";
@@ -44,13 +44,14 @@ export const IconWrapperTW = tw.span`
  * @property label - the text label
  * @property numCases - the number of cases in the cohort
  * @property filters - the filters to use for the cohort
- * @property filtersCallback - a callback to create the filters
+ * @property createStaticCohort - whether to create a case set from the filters so the cases in the cohort remain static
  */
 interface CohortCreationButtonProps {
   readonly label: ReactNode;
   readonly numCases: number;
-  readonly filters?: FilterSet;
-  readonly filtersCallback?: () => Promise<FilterSet>;
+  readonly filters: FilterSet;
+  readonly caseFilters?: FilterSet;
+  readonly createStaticCohort?: boolean;
 }
 
 /**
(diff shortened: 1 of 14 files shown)


wteouchicago:
## Issues
1. Not really sure if it's an issue, just weird and doesn't seem to happen in develop. In Cohort Comparison, when I create a cohort off the clinical tables, I see many more graphql calls in the network tab than expected. Steps:
- make a brain cohort based on home page link
- make a lung cohort based on home page link
- go to cohort comparison (my brain cohort is S1)
- create a cohort off S1 for male in gender (787 cases for me)
- in the network tab, I see:
-- a graphql call for creating the set of 787 cases
-- cohort creation call
-- countsQuery for the case counts on the tool cards. 
-- then 4 calls to create sets(?) these seem related to the survival plot table cases. Are these creating more sets than we want? why are they frozen? why do I have so many questions?


2. Getting some unexpected counts when creating a cohort from **context-sensitive** gene summary page's cancer distribution. Steps:
- Lung cohort from home page
- Mutation Frequency
- select VEP = low on left
- click on top gene - MUC16
- in cancer distribution table, use the create cohort button for TCGA-LUSC CNV Gains (21 cases) and create a cohort
- then create a cohort off **CNV Loss** for TCGA-LUSC (29 cases)
-- in the network tab, the graphql call to make the set has **cnv_change = gain** as filter
- if you look at the CNV Loss cohort, it'll have 21 cases instead of 29

But be aware of https://gdc-ctds.atlassian.net/browse/PEAR-1649 , which we should hopefully fix soon.

3. Some unexpected counts creating cohorts off the **context-insensitive** gene page's cancer distribution table. Steps:
- go to TP53's page: https://localhost.gdc.cancer.gov:3010/genes/ENSG00000141510
- for the CNV loss and CNV gain cohorts for TCGA-OV, they contain all the cases in the GDC (no filters)
-- the SSM Affected Cases cohort is OK

amlehman:
Hi @wteouchicago, 
I fixed 2 & 3. Number 1 is kind of ... complicated. I made some changes to move the set creation to the point where a user actually saves the cohort but the survival plot has special complicated filters that can't be represented by the filter representation we're using in the button so I moved the set creation into the Survival card component. I'll probably just go ahead and move it back into the modal, that will result in less calls (also the sets should be ephemeral, not frozen, whoops) 😞

wteouchicago:
looks good 👍 issues previously noted are fixed for me. 👍

packages/portal-proto/src/components/CohortCreationButton.tsx
```diff
@@ -44,28 +44,35 @@ export const IconWrapperTW = tw.span`
  * @property label - the text label
  * @property numCases - the number of cases in the cohort
  * @property filters - the filters to use for the cohort
- * @property filtersCallback - a callback to create the filters
+ * @property filtersCallback - callback to create filters, used when filters are too complicated for FilterSet
+ * @property createStaticCohort - whether to create a case set from the filters so the cases in the cohort remain static
```
paribartandhakal, line 49: Need to add definition for `caseFilters`.

packages/portal-proto/src/components/CohortCreationButton.tsx
```diff
@@ -44,28 +44,35 @@ export const IconWrapperTW = tw.span`
...
+  readonly createStaticCohort?: boolean;
 }
 
 /**
  * Button to create a new cohort
  * @param label - the text label
  * @param numCases - the number of cases in the cohort
  * @param filters - the filters to use for the cohort
- * @param filtersCallback - a callback to create the filters
+ * @param caseFilters - the case filters to use for the cohort
```
paribartandhakal, line 65: I see it here but I think we also need it above?

packages/portal-proto/src/features/set-operations/CountButtonWrapperForSetsAndCases.tsx
```diff
@@ -135,7 +135,6 @@ const CountButtonWrapperForSetsAndCases: React.FC<
 
   const createCohort = async () => {
     return await createSet({
-      // TODO: possibly add error handling
```
paribartandhakal, line 138: Is this done?

paribartandhakal (approved):

packages/portal-proto/src/features/set-operations/CountButtonWrapperForSetsAndCases.tsx
amlehman: No, but none of the other set creation buttons have error handling so its consistent lol. @wteouchicago, do we want to add any error handling for a failed set creation request on the cohort creation button?

craigrbarnes: We should. However, we need to be consistent about how we handle such errors. I propose we do this in a separate ticket.

wteouchicago: yeah, definitely a separate ticket :)

craigrbarnes (approved):
Changes are fine

PR #1125, opened 2024-05-16
PEAR-1960 - Investigate and fix: New filter cards not showing up sometimes? (Repository)

amlehman:
## Description
The issue was caused by combining the default facets and user custom facets into one and throwing everything into the store. Due to that, filters cards aren't getting updated because we are pulling the facet information from the persisted store. The fix I implemented was to only store a user's custom filters in the store and pull the default facet information from the JSON file so it will properly update when the JSON is updated. 

This only fixes the repository, I'll need to do something similar to fix cohort builder (mutation frequency and projects are fine, they don't follow this pattern).

To test:
1) On develop: try to update filters.json in the repository app and see that your change isn't reflected :( Add some custom filters to test that they will survive this change.
2) Switch over to this branch: update filters.json, see that your custom filters are still there.
 
## Checklist

- [x] Added proper unit tests
      flagged issues

### DIFF
--- a/packages/core/src/features/gdcapps/GdcApp.tsx
+++ b/packages/core/src/features/gdcapps/GdcApp.tsx
@@ -121,9 +121,7 @@ export const createAppID = (name: string, version: string): string =>
 // Apps with Local Storage
 //
 
-export const createAppStore = (
-  options: CreateGDCAppStore,
-): Record<any, any> => {
+export const createAppStore = (options: CreateGDCAppStore) => {
   const { name, version, reducers } = options;
   const nameVersion = `${name}::${version}`;
   const id = createAppID(name, version);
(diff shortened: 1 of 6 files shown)


packages/core/src/features/gdcapps/GdcApp.tsx
```diff
@@ -121,9 +121,7 @@ export const createAppID = (name: string, version: string): string =>
 // Apps with Local Storage
 //
 
-export const createAppStore = (
-  options: CreateGDCAppStore,
-): Record<any, any> => {
+export const createAppStore = (options: CreateGDCAppStore) => {
```
amlehman, line 124: Fixes typing for `useAppSelector`

packages/portal-proto/src/features/repositoryApp/appApi.ts
```diff
@@ -18,11 +19,25 @@ const downloadAppReducers = combineReducers({
...
 });
 
+const migrations = {
+  2: (state) => {
+    return {
+      ...state,
+      facets: {
+        customFacets: state.facets.facets.filter(
+          (facet) => !RepositoryDefaultConfig.facets.includes(facet),
+        ),
```
amlehman, line 29: This allows us to alter the store to only store the custom facets for users who already have a persisted store.

wteouchicago:
works well for me - very nice! 👍

craigrbarnes (approved):
Code are fine, and fixes the issue

PR #1134, opened 2024-05-24
PEAR-1688: remove score when creating gene or ssm set

craigrbarnes:
## Description
Fixes an issue where gene/ssm sets are size 0. Adds case_filters to the gene/ssm setCreation query. I removed the score from these queries and added `useCreateTopNGeneSetFromFiltersMutation` and `useCreateTopNSsmsSetFromFiltersMutation`. Both functions use the passed filter to get a list of gene/ssms ids used to create the set. 

There is also a fix for failing unit test for ColumnOrdering.

Testing note:

One test for this is:

1.  Open Cohort Builder:
2.  Create a cohort with` files.experimental_strategy = WXS`
3.  Open Mutation Frequency
4.  Create Save/Edit Gene Set -> Save as new gene set (with no genes selected) 
5.  The created set should have > 0 genes
also fixes: https://gdc-ctds.atlassian.net/browse/SV-2314

## Checklist

- [ ] Added proper unit tests
      flagged issues

### DIFF
--- a/packages/portal-proto/src/components/Table/test/ColumnOrdering.unit.test.tsx
+++ b/packages/portal-proto/src/components/Table/test/ColumnOrdering.unit.test.tsx
@@ -70,7 +70,7 @@ describe("ColumnOrdering", () => {
       screen.getByTestId("column-selector-popover-modal"),
     ).toBeInTheDocument();
 
-    await userEvent.click(button);
+    await userEvent.click(button, { delay: 100 });
     await waitFor(() =>
       expect(screen.queryByTestId("column-selector-popover-modal")).toBeNull(),
     );
--- a/packages/portal-proto/src/components/Table/test/ColumnOrdering.unit.test.tsx
+++ b/packages/portal-proto/src/components/Table/test/ColumnOrdering.unit.test.tsx
@@ -49,7 +49,7 @@ describe("ColumnOrdering", () => {
       />,
     );
     const button = screen.getByTestId("button-column-selector-box");
-    await userEvent.click(button);
+    await userEvent.click(button, { delay: 100 });
     expect(
       screen.getByTestId("column-selector-popover-modal"),
     ).toBeInTheDocument();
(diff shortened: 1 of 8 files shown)


packages/portal-proto/src/components/Table/test/ColumnOrdering.unit.test.tsx
```diff
@@ -65,7 +65,7 @@ describe("ColumnOrdering", () => {
       />,
     );
     const button = screen.getByTestId("button-column-selector-box");
-    await userEvent.click(button);
+    await userEvent.click(button, { delay: 100 });
```
paribartandhakal, line 68: We might need in line 73 also?

craigrbarnes: Agreed. I will add it.

paribartandhakal: Actually, I found `await userEvent.click(button)` in more places too. I am not sure what changed in mantine 7 to fail these tests. Do you have any idea?

wteouchicago:
this worked well for me. 👍 there is some slowness when saving a ton of top X SSMs due to response times from API, but that's expected for now.

packages/portal-proto/src/components/Table/test/ColumnOrdering.unit.test.tsx
craigrbarnes: I do. For dropdowns, the Popup Modal is not part of the DOM unit button is pressed, so the delay seems to allow the DOM to update. Since only this test is failing, maybe just fix it where it's failing?

paribartandhakal: Yes makes sense.

paribartandhakal (approved):

PR #1136, opened 2024-05-30
PEAR-1787: Update data after the user logs in OR out

paribartandhakal:
## Description
Injecting `loggedIn` state before returning filters which in turn will make the query arg change resulting in rtk query refetching the cached API calls.

## Checklist
- [ ] Added proper unit tests
      flagged issues


wteouchicago:
I tested pre-PR and it worked great for me. 👍

packages/core/src/features/gdcapi/filters.ts
```diff
@@ -260,11 +260,13 @@ export interface GqlExcludeIfAny {
 export interface GqlIntersection {
   readonly op: "and";
   readonly content: ReadonlyArray<GqlOperation>;
+  readonly isLoggedIn?: boolean;
```
amlehman, line 263: Is this actually getting used? Why do we need this state both on the cohort and in the filters?

packages/core/src/features/cohort/availableCohortsSlice.ts
```diff
@@ -582,6 +587,14 @@ export const selectCurrentCohortId = (state: CoreState): string | undefined =>
 export const selectCohortMessage = (state: CoreState): string[] | undefined =>
   state.cohort.availableCohorts.message;
 
+/**
+ * Returns the current cohort login status
```
amlehman, line 591: Can we expand this comment and explain the purpose of these logged in flags are? Its not very obvious why this value would be associated with cohort/filters.

amlehman (approved):

craigrbarnes (approved):
Changes are fine and resolved the issues in the ticket.

packages/core/src/features/cohort/availableCohortsSlice.ts
craigrbarnes: +1

paribartandhakal: I will update it. Good call.

paribartandhakal:
> Changes are fine and resolved the issues in the ticket.

Thank you for resolving the conflicts.

packages/core/src/features/cohort/availableCohortsSlice.ts
craigrbarnes: Something like: 

The case composition of a cohort can be different based on the login status of the user. The login status is attached to the cohort so that the portal and applications are updated are when the login state changes. 

Feel free to use any or none of this.

paribartandhakal: This is much better than what I came up with. Too bad I didn't notice in time. I will still update it.

paribartandhakal: Okay this has been updated. Thank you 😃

PR #1119, opened 2024-05-13
PEAR-1852: Fix for cDAVE - Survival plot resets what is plotted when switching from days to years (or vice-versa)

paribartandhakal:
## Checklist
- [N/A] Added proper unit tests
- [N/A] Left proper TODO messages for any remaining tasks
- [N/A] Scanned for web accessibility with **aXe**, and mitigated or documented
      flagged issues

PR #1144, opened 2024-06-03
PEAR-1974: Fix updating survival plot in CDave issue

paribartandhakal:
## Description
Added condition to check before updating survival plots.

## Checklist
      flagged issues

### DIFF
--- a/packages/portal-proto/src/features/cDave/CDaveCard/ContinuousData.tsx
+++ b/packages/portal-proto/src/features/cDave/CDaveCard/ContinuousData.tsx
@@ -1,4 +1,4 @@
-import { useState, useMemo } from "react";
+import { useRef, useState } from "react";
 import { useDeepCompareEffect, useDeepCompareMemo } from "use-deep-compare";
 import {
   useGetContinuousDataStatsQuery,
@@ -100,6 +100,7 @@ const ContinuousData: React.FC<ContinuousDataProps> = ({
   );
   const [selectedFacets, setSelectedFacets] = useState<SelectedFacet[]>([]);
   const [yTotal, setYTotal] = useState(0);
+  const dataDimensionRef = useRef(dataDimension);
 
   const ranges = useDeepCompareMemo(
     () =>
@@ -138,7 +139,7 @@ const ContinuousData: React.FC<ContinuousDataProps> = ({
     },
   });
 
-  const displayedData = useMemo(
+  const displayedData = useDeepCompareMemo(
     () =>
       processContinuousResultData(
         isSuccess ? data : {},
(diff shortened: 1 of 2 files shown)


packages/portal-proto/src/features/cDave/CDaveCard/ContinuousData.tsx
```diff
@@ -161,13 +164,14 @@ const ContinuousData: React.FC<ContinuousDataProps> = ({
...
       );
+    }
 
     if (customBinnedData === null) {
       setYTotal(displayedData.reduce((a, b) => a + b.count, 0));
     }
 
     setSelectedFacets([]);
-  }, [displayedData, customBinnedData]);
+  }, [dataDimension, displayedData, customBinnedData]);
```
amlehman, line 174: Couldn't you just add dataDimension to the dependency array 🤔? Why is the ref necessary?

paribartandhakal: We don't want to update the survival plot when the dataDimension changes. So I am using ref to check that. 

If we just add `dataDimension` to the dep array, it will trigger the survival plot changes which we don't want.

Does this make sense? Also, do you think there's a better way to achieve this?

amlehman: Ah, I see. I think this solution is a little confusing and prone to resulting in bugs where the things don't update when we expect them to. `dataDimension` represents whether the card is displaying years vs days, so I'm not sure how I get how this is fixing the issue described in the ticket. I think we probably need a useEffect looking for changes on customBinned data that resets selectedSurvivalPlots to an empty array.

amlehman: Ah, I see. I think this solution is a little confusing and prone to resulting in bugs where the things don't update when we expect them to. `dataDimension` represents whether the card is displaying years vs days, so I'm not sure how I get how this is fixing the issue described in the ticket. I think we probably need a useEffect looking for changes on customBinned data that resets selectedSurvivalPlots to an empty array.

paribartandhakal: I see how this might be confusing. I previously [worked](https://github.com/NCI-GDC/gdc-frontend-framework/pull/1119) on a fix for this [ticket](https://gdc-ctds.atlassian.net/browse/PEAR-1852) to prevent the survival plot from updating when dataDimension changes. However, my earlier changes introduced a new issue reported in `PEAR-1974`. To address this, I'm proposing a new solution for `PEAR-1852` that avoids unintended side effects.

wteouchicago:
works for me. 👍

amlehman (approved):

craigrbarnes (approved):

PR #1151, opened 2024-06-05
PEAR-1964 -  Age at Index is displaying incorrect values in Cohort Builder

amlehman:
## Checklist

- [x] Added proper unit tests
      flagged issues

### DIFF
--- a/packages/core/src/features/facets/facetDictionaryApi.ts
+++ b/packages/core/src/features/facets/facetDictionaryApi.ts
@@ -75,9 +75,17 @@ export const classifyFacetDatatype = (f: FacetDefinition): FacetTypes => {
   if (fieldName.includes("age_is_")) return "enum";
   if (fieldName.includes("datetime")) return "datetime";
   if (fieldName.includes("percent")) return "percent";
-  if (fieldName.includes(".age_")) return "age";
-  if (fieldName.includes("_age_")) return "age";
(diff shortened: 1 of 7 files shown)


wteouchicago:
### Issue
1. In develop as well. May be related to the original bug that this PR is trying to fix. For Age At Index, when we enter values as days (e.g. 0-10), it's querying with these values so I end up filtering for 0-10 years. It should be filtering for (0 - 10 days converted  to years, rounded, etc.), which is kind of 🤮 , frankly, but would be consistent with cards with Days-Years toggle.
- I'm tempted to just remove the Days tab for cards whose values are already in Years. This would make it consistent with cDAVE's Age at Index card.
- qn: which change is easier? It doesn't need to be in this PR.

____

looks good otherwise.

amlehman:
@wteouchicago, I think removing the day option would be easier. The facet day/year logic has gotten to be pretty tangled imo. I think it could probably use some reworking before attempting to convert values in another place.

wteouchicago:
OK! I'll create a separate ticket for removing the days option. 👍 

changes here look good. 👍

paribartandhakal (approved):

craigrbarnes:
Agree this needs some untangling. If we ever get the dictionary mapping  API we can simplify this even more.

PR #1170, opened 2024-06-12
PEAR-2005 -  Set Operations - Saving an unsaved cohort causes "Unexpected Error" modal to appear

amlehman:
## Description
This was caused by two issues:
1) the set operations app was not updating correctly when cohorts changed (SV-2474). Pari updated the set operations app to update correctly with PEAR-1787, revealing the second issue. 
2) The id of the cohort changes when it is saved because we receive an id back from the API that we need to refer to the cohort by. Fix is to keep the old unsaved cohort id around so the set operation app can find the cohort associated with the id that the user initially selected. 

## Checklist

- [x] Added proper unit tests
      flagged issues


wteouchicago:
changed base to 2.14

wteouchicago:
looks good 👍

craigrbarnes (approved):
Changes look fine

PR #1181, opened 2024-06-18
PEAR-1963: cDAVE - Search input should remain fixed and not scroll

paribartandhakal:
## Description
1. Style Changes
2. Code Clean Up

@wteouchicago please verify the overall behavior of CDave specially the fields panel on the left. Thank you 🙇‍♂️ 

## Checklist
- [ ] Added proper unit tests
      flagged issues

https://github.com/NCI-GDC/gdc-frontend-framework/assets/101295912/f3cad20e-e66e-4367-af7b-d7d116e02cc2

### DIFF
--- a/packages/portal-proto/src/features/cDave/Controls.tsx
+++ b/packages/portal-proto/src/features/cDave/Controls.tsx
@@ -80,9 +80,9 @@ const ControlGroup: React.FC<ControlGroupProps> = ({
   return filteredFields.length > 0 ? (
     <>
       <span
-        onClick={() => setGroupOpen((prev) => !prev)}
+        onClick={() => setGroupOpen(!groupOpen)}
         onKeyDown={createKeyboardAccessibleFunction(() =>
-          setGroupOpen((prev) => !prev),
+          setGroupOpen(!groupOpen),
         )}
         tabIndex={0}
         role="button"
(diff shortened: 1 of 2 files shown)


wteouchicago:
The scroll behaviour looks right to me -- cemile would be the best judge and we can have her look after merge. I also sanity-checked cDAVE in general, but not every nook and cranny of functionality. Let me know if there's something I should look out for.

One thing -- do you think we need some padding on the right side of the search box? so that it's rightmost edge is aligned with the rightmost edge of the blue section headers (e.g. Demographic)?

craigrbarnes (changes requested):
Agree with @wteouchicago to add right padding to the search input for better alignment with the facet header's right section.

Some issues with the less button:

1. The size of the + button is now different than the - button:

image

image

2. Clicking on the less button (the minus icon) does not close the expanded list. Clicking on the "show less" text does work  as before

paribartandhakal:
> Agree with @wteouchicago to add right padding to the search input for better alignment with the facet header's right section.
> 
> Some issues with the less button:
> 
> 1. The size of the + button is now different than the - button:
> 
> image
> 
> image
> 
> 2. Clicking on the less button (the minus icon) does not close the expanded list. Clicking on the "show less" text does work  as before

Good call. I fixed both the issues and fixed a failing unit test.

paribartandhakal:
> The scroll behaviour looks right to me -- cemile would be the best judge and we can have her look after merge. I also sanity-checked cDAVE in general, but not every nook and cranny of functionality. Let me know if there's something I should look out for.
> 
> One thing -- do you think we need some padding on the right side of the search box? so that it's rightmost edge is aligned with the rightmost edge of the blue section headers (e.g. Demographic)?

I updated the styling.

craigrbarnes:
Items 1 and 2 are addressed.  Everything else looks fine.

craigrbarnes (approved):

PR #1198, opened 2024-07-03
PEAR-1379 - Cohort Builder - Inconsistent behavior when removing filters 

amlehman:
## Description
Value for NumberInput needs to be an empty string instead of undefined to clear out the value according to mantine 🤷‍♀️ https://github.com/mantinedev/mantine/issues/6483

## Checklist

- [x] Added proper unit tests
      flagged issues

### DIFF
--- a/packages/portal-proto/src/features/facets/NumericRangeFacet.tsx
+++ b/packages/portal-proto/src/features/facets/NumericRangeFacet.tsx
@@ -1,4 +1,5 @@
 import React, { useCallback, useEffect, useMemo, useState } from "react";
+import { useDeepCompareMemo } from "use-deep-compare";
 import { useDeepCompareEffect } from "react-use";
 import {
   MdClose as CloseIcon,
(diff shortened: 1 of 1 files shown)


packages/portal-proto/src/features/facets/NumericRangeFacet.tsx
```diff
@@ -782,12 +788,21 @@ const NumericRangePanel: React.FC<NumericFacetData> = ({
 }: NumericFacetData) => {
   const adjMinimum = minimum != undefined ? minimum : 0;
   const adjMaximum = maximum != undefined ? maximum : 999999;
+
+  const filter = hooks.useGetFacetFilters(field);
+  const [filterValues] = useMemo(() => {
```
paribartandhakal: might wanna use `useDeepcompareMemo` given filter is an object or does it not matter here?

wteouchicago:
looks great! 👍

packages/portal-proto/src/features/facets/NumericRangeFacet.tsx
craigrbarnes: Using useDeepcompareMemo might be better. The selector itself is not memorized so this potentially would mitigate it.

paribartandhakal (approved):

PR #1200, opened 2024-07-05
DR test annotations QA-2200

bpierce01:
## Description
- Test for annotations added in dr-40


ManjunathJayanna (approved):
Quick test, Approving.

PR #1226, opened 2024-08-05
PEAR-1994 - Cohort Builder, remove years-days toggle for age at index

amlehman:
## Description
\+ Fix the bug in the ticket with the inputs not displaying your selected range
\+ Fix newly discovered bug where custom filters weren't correctly populating in tab on page load

## Checklist

- [x] Added proper unit tests
      flagged issues


wteouchicago:
## Issues
1. Something odd has happened to the Year of Diagnosis card (General Diagnosis category). It almost looks like it's acting like the original values are in days, but are converted to years and displaying only years (plus a weird bucketing starting at 5 years). See screenshot - left is develop, right is this branch.
- the same is happening with Tobacco Smoking Onset Year (in Exposure)
- guessing this is happening to the calendar years type cards
 
Screenshot 2024-08-05 at 2 30 16 PM

2. Just a possible issue. I'm discussing with the leads how we should handle it and will circle back. In this branch, I see the bucket 90-100, which is correct since we do have some cases that have '90' as the age of index. The max for this card, however, seems to be set at 89 for the custom fields in this branch. It's 90 on develop (guessing due to it being lumped with the types of fields that have days-years toggle?). ~~Qn: can the max be changed for this card to 90 (no need to do this yet)?~~
- Edit: let's not display the 90-100 bucket (Craig says we can remove the 90-100 from the query?).

packages/portal-proto/src/features/facets/NumericRangeFacet.tsx
```diff
@@ -741,7 +751,7 @@ const Year: React.FC<NumericFacetData> = ({
     <div className="flex flex-col w-100 space-y-2 px-2 mt-1 ">
       <RangeInputWithPrefixedRanges
         hooks={{ ...hooks }}
-        units="year"
+        units="years"
```
craigrbarnes, line 754: Note: There is a year type, which is a calendar year, vs. years, which are actually filtered by days.

craigrbarnes: I deleted this, but have determined that this is the cause of the issue @wteouchicago reported. There is a unit type **year**, which represents a calendar year, as opposed to **years**, which is converted to days.

craigrbarnes:
> ## Issues
> 1. Something odd has happened to the Year of Diagnosis card (General Diagnosis category). It almost looks like it's acting like the original values are in days, but are converted to years and displaying only years (plus a weird bucketing starting at 5 years). See screenshot - left is develop, right is this branch.
> 
> * the same is happening with Tobacco Smoking Onset Year (in Exposure)
> * guessing this is happening to the calendar years type cards
> 
> Screenshot 2024-08-05 at 2 30 16 PM
> 2. Just a possible issue. I'm discussing with the leads how we should handle it and will circle back. In this branch, I see the bucket 90-100, which is correct since we do have some cases that have '90' as the age of index. The max for this card, however, seems to be set at 89 for the custom fields in this branch. It's 90 on develop (guessing due to it being lumped with the types of fields that have days-years toggle?). ~Qn: can the max be changed for this card to 90 (no need to do this yet)?~
> 
> * Edit: let's not display the 90-100 bucket (Craig says we can remove the 90-100 from the query?).

For age at index, it looks like the min/max values are passed as days when it's expecting the value in years. If you fix this, the 90-100 busk should disappear. Note that the query filters have 3251 buckets. Also, I believe the age of index is stored in years not days, which explains why the values of the ranges are correct even though it's asking for 3251 buckets from 0 to 32507

packages/portal-proto/src/features/facets/NumericRangeFacet.tsx
amlehman: Reverted this

craigrbarnes:
Age of index still seems to have an issue: A custom range results in >= 0  AND < 0 

image

wteouchicago:
looks good - not spotting anything unusual! 👍

craigrbarnes (approved):
Changes are fine and works correctly

PR #1235, opened 2024-08-09
(PEAR-2098): Properly check for undefined values in range card

ocshawn:
## Description
- Properly check for undefined values in range card so it displays 0 values 

## Checklist

- [x] Added proper unit tests
      flagged issues


packages/portal-proto/src/features/facets/NumericRangeFacet/FromTo.tsx
```diff
@@ -251,7 +253,7 @@ const FromTo: React.FC<FromToProps> = ({
             className="text-sm flex-1"
             placeholder={`Min: ${lowerUnitRange}${unitsLabel} `}
             // units are always days
-            value={form.values.fromValue ?? ""}
```
amlehman, line 273: I think these need to stay as empty strings or it will regress PEAR-1379: https://github.com/NCI-GDC/gdc-frontend-framework/pull/1198

ocshawn: good catch, changed

amlehman (approved):

wteouchicago:
When I enter "0" in the From: and "1" in the To: for Age of Diagnosis, then click Apply, the "0" is removed and I see just the "Min: -90 years". It looks OK in the query expressions section though.

If I enter "0" in the From: again and click Apply, this time the "0" sticks around. 

The same happens in the To: field, and also to other range cards.

ocshawn:
looked into the mantine form documentation and changed our implementation to better match, let me know if you see any bugs related to undefined

ocshawn:
found another bug continuing to look into

ocshawn:
should be ready for review now

wteouchicago:
### Issues
1. This seems to reintroduce the problem that Amy's previous PR was fixing -- when we removing values from filters in the Query Expressions, the associated fields in the filter cards are not cleared: https://github.com/NCI-GDC/gdc-frontend-framework/pull/1198

Example:
- use Age at Diagnosis card in Demographic
- type "1" for From, "2" for To; Apply
- in the Query Expressions (QE) section, click on one of the operators to remove the associated values (there seems to be a separate bug here that's also in develop and prod where the operator you are clicking on is the one that remains rather than the one that is removed-- I'll file separately).
- note that in the Age at Diagnosis card, the value removed from the QE section is still displayed in the field

2. the Clear Selection button for the range cards no longer clears the values in the fields if you had typed in some numbers previously. (It works if you had previously just selected one of the predefined buckets). 

3. similar to 2 above. If I do some filtering, then remove the entire property in the QE section, the numbers remain in the fields. Actually, sometimes the numbers that are still in the field are not what I had entered, e.g. go to General Diagnosis > Year of Diagnosis card. Enter 1900 in From: and 2050 in To. Remove the property in the QE section. I end up with "1" and "2" in the fields.

craigrbarnes (approved):

wteouchicago (dismissed):
see previous comment

wteouchicago:
Looks good! Didn't spot anything unusual. 👍

PR #1237, opened 2024-08-12
(PEAR-2099): Warning Modal dismiss text change

ocshawn:
## Description
- Warning Modal that appears when user has no controlled access the dismiss button should say "Close"  

## Checklist

- [x] Added proper unit tests
      flagged issues

### DIFF
--- a/packages/portal-proto/src/components/Modals/UserProfileModal.tsx
+++ b/packages/portal-proto/src/components/Modals/UserProfileModal.tsx
@@ -72,7 +72,7 @@ export const UserProfileModal = ({
       }
       openModal={openModal}
       size="60%"
-      buttons={[{ title: "Done", dataTestId: "button-user-profile-done" }]}
+      buttons={[{ title: "Close", dataTestId: "button-user-profile-done" }]}
     >
       <div className={`${!data ? "py-4" : "py-2"}`}>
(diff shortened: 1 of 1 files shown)


packages/portal-proto/src/components/Modals/UserProfileModal.tsx
```diff
@@ -72,7 +72,7 @@ export const UserProfileModal = ({
       }
       openModal={openModal}
       size="60%"
-      buttons={[{ title: "Done", dataTestId: "button-user-profile-done" }]}
+      buttons={[{ title: "Close", dataTestId: "button-user-profile-done" }]}
```
paribartandhakal, line 75: I feel like this might not be correct because the modal shows two different things. Can we say `Done` for when the data length is greater than 0 and `Close` for the other case?

cc. @wteouchicago

wteouchicago:
I think "Close" works for all the situations if I'm understanding correctly:
- the warning modal when you don't have any controlled access and you try to download a token
- the user profile when you have controlled access to >= 1 project
- the user profile when you don't have controlled access to projects

("Done" was just really weird when we are telling people they don't have access.)

paribartandhakal (approved):

craigrbarnes (approved):

PR #1239, opened 2024-08-12
PEAR-748: Add spinners when loading or redrawing

paribartandhakal:
## Description
Added loading overlay in appropriate places.

## Checklist
- [ ] Added proper unit tests
      flagged issues


wteouchicago:
1. For the spinner on submit button for filtering your cohort by cases/genes/mutations,  may want to make it consistent with what we do when we import new cohort or save cohort (with naming modal) --> it seems the spinner appear to the left of the button label there.  i don't have a strong preference for which version we pick, just prefer it consistent. Possibly just making the spinner show to the left of the label is better since existing Submit spinners already behave that way.
- same for the saving genes/mutations sets from MF

2. I'm still seeing NaNs for the venn diagram in set operations while things are loading

paribartandhakal:
> 1. For the spinner on submit button for filtering your cohort by cases/genes/mutations,  may want to make it consistent with what we do when we import new cohort or save cohort (with naming modal) --> it seems the spinner appear to the left of the button label there.  i don't have a strong preference for which version we pick, just prefer it consistent. Possibly just making the spinner show to the left of the label is better since existing Submit spinners already behave that way.
> 
> * same for the saving genes/mutations sets from MF
> 
> 2. I'm still seeing NaNs for the venn diagram in set operations while things are loading

I just fixed them. Let me know if there are any other issues.

packages/portal-proto/src/components/Modals/SetModals/SaveSelectionModal.tsx
```diff
@@ -165,9 +166,12 @@ const SaveSelectionAsSetModal: React.FC<SaveSelectionAsSetModalProps> = ({
                   color: "red",
                   closeButtonProps: { "aria-label": "Close notification" },
                 });
-              })
+              });
+          }}
+          disabled={!form.isValid()}
+          leftSection={
+            response?.isLoading && <Loader size="sm" color="white" />
```
amlehman: For typing reasons, I think its generally best to avoid this way of rendering when passing in a component as a prop. When the condition is false, you're passing in the value of false instead of undefined to `leftSection`. Depending on the how the client library is written, it could be a type error or just behave in a way you're not expecting. Looks like mantine has this prop typed as `ReactNode` so a boolean is fine but if they choose `ReactElement` then false would be a type error where undefined is always 'safe'.

amlehman (approved):

packages/portal-proto/src/components/Modals/SetModals/SaveSelectionModal.tsx
paribartandhakal: should I use ternary and pass in `null`?

amlehman: Ternary and undefined

wteouchicago:
Issue 1 looks good.
For issue 2, very minor -- is it easy to show "..." instead of "0" when the venn diagram is loading? no need to implement if it's not trivial -- "0" is better than the "NaN".

paribartandhakal:
> Issue 1 looks good. For issue 2, very minor -- is it easy to show "..." instead of "0" when the venn diagram is loading? no need to implement if it's not trivial -- "0" is better than the "NaN".

updated

packages/portal-proto/src/components/Modals/SetModals/SaveSelectionModal.tsx
paribartandhakal: updated

wteouchicago:
The NaNs are back :P 

Screenshot 2024-08-14 at 10 31 59 AM

wteouchicago:
The NaNs are gone! 👍

craigrbarnes (approved):

PR #1241, opened 2024-08-13
PEAR-1971 - Add error handling for a failed set creation request on the cohort creation button

amlehman:
## Description
The easiest way to test this is to set up whatever set action you want to test and switch to offline in the throttling options on the dev console, ie:

https://github.com/user-attachments/assets/3613f2f1-f551-49b4-ae3d-e8db2a3b5e4c

## Checklist

- [x] Added proper unit tests
      flagged issues


packages/core/src/features/gdcapi/gdcgraphql.ts
```diff
@@ -84,7 +90,11 @@ export const graphqlAPISlice = coreCreateApi({
     try {
       results = await graphqlAPI(request.graphQLQuery, request.graphQLFilters);
     } catch (e) {
-      return { error: e as GraphQLFetchError };
+      return {
+        error: errorIsGraphQLError(e as Error | GraphQLFetchError)
```
amlehman, line 94: This is to fix an issue where network failures were adding unserializable objects to the store.

wteouchicago:
This worked for me. But I'm thinking that if there is an error saving/creating/export something, we should be showing a modal rather than a notification (more toasts, more problems?). Question:
- would you rather have a separate ticket tackling all the notifications that really should be modals?
- or do you want to change the new notifications we've added to modals here?

For the modal, I'm thinking something similar to the "Save Cohort Error" modal. Something like these messages could work if we want to do it with this PR (and I can adjust more later too).

"SAVE COHORT ERROR"
- There was a problem saving the cohort. (For this, I don't think users need to distinguish too much whether it's at the creating or saving step that things broke.)

"EXPORT SET ERROR"
- There was a problem exporting the set.

amlehman:
Hi @wteouchicago, I updated the error messages. Let me know if everything looks good.

wteouchicago:
Changes look good. I realise there is one more area that is currently a toast but could be a modal (again, lemme know if you prefer a new ticket and doing it separately): when there is some error saving sets (e.g. from genes/mutations table, and from set operations when sets, instead of cohorts, are involved). A message like the following could work:

"SAVE SET ERROR"
- There was a problem saving the set.

amlehman:
Hi @wteouchicago, sorry kind of forgot about this ticket 😓 I've made that update.

craigrbarnes (approved):

wteouchicago:
@amlehman no worries, as long as you are not mad about the other places I just found ;) All of them can have the same "SAVE SET ERROR", "There was a problem saving the set." modal instead.

1. In the genes and mutations tables, the 2nd and 3rd option in the "Save/Edit...Set" dropdown (Add to... / Remove from...). When there is an error saving, a notification with "Problem modifying set" is shown. 

2. In Manage Sets > Create Gene/Mutation Set > Submit

3. Possibly related to (2):
- Mutation Frequency > Upload Genes/Somatic Mutations > Save Set
- Cohort Builder > Molecular Filters > Upload Genes/Somatic Mutations > Save Set
- Cohort Bar > Custom Filters > Mutated Genes / Somatic Mutations > Save Set

PR #1244, opened 2024-08-13
PEAR-2028 - Improve Quick Search even more 

amlehman:
## Description
Screenshot 2024-08-13 at 3 00 47 PM

## Checklist

- [x] Added proper unit tests
      flagged issues

### DIFF
--- a/packages/portal-proto/src/components/QuickSearch/tests/QuickSearch.unit.test.tsx
+++ b/packages/portal-proto/src/components/QuickSearch/tests/QuickSearch.unit.test.tsx
@@ -97,4 +97,34 @@ describe("<QuickSearch />", () => {
     const noResults = getByTestId("no-results-quick-search-bar");
     expect(noResults).toBeInTheDocument();
   });
+
+  test("displays no results found if file doesn't have a newer version", async () => {
+    jest.spyOn(core, "useQuickSearchQuery").mockReturnValue({
+      data: {
+        searchList: [],
+        query: "111-222",
+      },
+    } as any);
+    jest.spyOn(core, "useGetHistoryQuery").mockReturnValue({
+      data: [
+        {
+          file_change: "released",
+          uuid: "111-222",
+          version: "1",
+        },
+      ],
+    } as any);
+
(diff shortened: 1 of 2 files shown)


wteouchicago:
works great! 👍

packages/portal-proto/src/components/QuickSearch/tests/QuickSearch.unit.test.tsx
```diff
@@ -97,4 +97,34 @@ describe("<QuickSearch />", () => {
...
+    userEvent.click(getByTestId("textbox-quick-search-bar"));
+    userEvent.type(getByTestId("textbox-quick-search-bar"), "111-222");
+
+    await waitFor(
+      () =>
+        expect(getByTestId("no-results-quick-search-bar")).toBeInTheDocument(),
+      { timeout: 3000 },
+    );
+    const noResults = getByTestId("no-results-quick-search-bar");
+    expect(noResults).toBeInTheDocument();
```
paribartandhakal, line 128: Is it doing the same thing?

amlehman: It's waiting to make sure the element is in the DOM (since its a dropdown that appears after the search is conducted) until it does the expect for the test. I'm not sure if there's a more elegant way to do that 🤔

paribartandhakal: I meant line 124 and 128 is testing for the same test-id. Is the latter still needed?

amlehman: I mean, I guess its not technically needed because the test will pass if nothing fails it but I just like to have some sort of positive/negative assertion in a test instead of relying on line 124 throw an exception to fail it.

craigrbarnes: I'm fine with that

craigrbarnes (approved):

PR #1251, opened 2024-08-22
Qa 2279 add tests dr 41

bpierce01:
## Description
- Added relevant tests based on DR-41 additions and removals 
- Updated statistic count test data for home page test
- Corrected test language in some DR-40 tests

Execution Instrutions:
VPN2, qa-yellow
gauge run specs/gdc_data_portal_v2/data_release/41/

### DIFF
--- a/holmes-py/specs/gdc_data_portal_v2/data_release/40/annotation_inclusion.spec
+++ b/holmes-py/specs/gdc_data_portal_v2/data_release/40/annotation_inclusion.spec
@@ -1,4 +1,4 @@
-# Data Release: 40 - Mutation Inclusion
+# Data Release: 40 - Annotation Inclusion
 Date Created        : 07/05/2024
 Version			    : 1.0
 Owner		        : GDC QA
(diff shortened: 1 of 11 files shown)


uc-pperumala:
The tests look good.
However, why not take advantage of data-driven execution for DRs where only the data changes dr over dr. Why don't we use a csv file that will hold the test data dr over dr, and in the spec files we can have the spec files run through the scenarios in the csv.
See https://docs.gauge.org/execution?os=macos&language=python&ide=vscode

bpierce01:
> The tests look good. However, why not take advantage of data-driven execution for DRs where only the data changes dr over dr. Why don't we use a csv file that will hold the test data dr over dr, and in the spec files we can have the spec files run through the scenarios in the csv. See https://docs.gauge.org/execution?os=macos&language=python&ide=vscode

Sure that makes sense, rather than repeating similar spec files. I’d like to do it in a separate ticket however, as I think we’d need to make a separate series of test steps that take in a csv file as opposed to direct input. I'll make a follow-up ticket on the QA board. 

Edit: https://gdc-ctds.atlassian.net/browse/QA-2283

uc-pperumala (approved):
Sounds good, thank you!

craigrbarnes (approved):

PR #1264, opened 2024-08-30
Qa 2281 create case table automation ids

bpierce01:
## Description
- Added automation IDs to cases summary view and cases table view

### DIFF
--- a/packages/portal-proto/src/components/PopupIconButton/PopupIconButton.tsx
+++ b/packages/portal-proto/src/components/PopupIconButton/PopupIconButton.tsx
@@ -26,7 +26,7 @@ export const PopupIconButton = forwardRef<
   ): JSX.Element => {
     return (
       <Button
-        data-testid={`link-open-${label}`}
+        data-testid={`Link to open ${label}`}
         className={customStyle}
         onClick={handleClick}
         aria-label={customAriaLabel ?? `Open ${label} information in modal`}
--- a/packages/portal-proto/src/components/PopupIconButton/PopupIconButton.tsx
+++ b/packages/portal-proto/src/components/PopupIconButton/PopupIconButton.tsx
@@ -26,6 +26,7 @@ export const PopupIconButton = forwardRef<
   ): JSX.Element => {
     return (
       <Button
+        data-testid={`link-open-${label}`}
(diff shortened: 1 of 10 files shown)


ManjunathJayanna (approved):

packages/portal-proto/src/components/Table/TotalItem.tsx
```diff
@@ -32,7 +32,8 @@ const TotalItems: React.FC<TotalItemsProps> = ({
 
   return (
     <>
-      Total of <b>{formattedTotal}</b> {itemText}
+      Total of <b data-testid="text-total-item-count">{formattedTotal}</b>{" "}
```
amlehman, line 35: Having the test id on this nested bold tag is a little strange to me 🤔. We could move it up to the vertical table component [here](https://github.com/NCI-GDC/gdc-frontend-framework/blob/develop/packages/portal-proto/src/components/Table/TableHeader.tsx#L95) but I'm not sure if you're trying to specifically select the table headers with item counts.

packages/portal-proto/src/components/PopupIconButton/PopupIconButton.tsx
```diff
@@ -26,6 +26,7 @@ export const PopupIconButton = forwardRef<
   ): JSX.Element => {
     return (
       <Button
+        data-testid={`Link to open ${label}`}
```
amlehman: Can this id be the same format (kebab case) as the rest of the test ids?

amlehman (approved):

packages/portal-proto/src/components/PopupIconButton/PopupIconButton.tsx
bpierce01: Done! Updated the id to be correct case as suggested

packages/portal-proto/src/components/Table/TotalItem.tsx
bpierce01: I put the ID there to have it directly on an item count we may check. If it doesn't hurt current or future formatting i'd like it there. Let me know

craigrbarnes (approved):

PR #1279, opened 2024-09-06
PEAR-2106: Cohort Builder - Demographic - Age at Diagnosis, fix the min

paribartandhakal:
## Checklist
- [X] Added proper unit tests
      flagged issues

age_at_diagnosis


packages/portal-proto/src/features/facets/NumericRangeFacet/types.ts
```diff
@@ -2,8 +2,8 @@ import { FacetCardProps, RangeFacetHooks } from "../types";
 
 export interface NumericFacetProps extends FacetCardProps<RangeFacetHooks> {
   readonly rangeDatatype: string;
-  readonly minimum?: number;
-  readonly maximum?: number;
+  readonly minimum: number;
```
amlehman: If these are required props, what is the case that the default value [here](https://github.com/NCI-GDC/gdc-frontend-framework/pull/1279/files#diff-b0e7154233efd3e562947e81932296bd4349b8971ad9818515bb22e9fd519cfbR40) gets used?

paribartandhakal: Some of the values are "undefined" for which I am setting a default value.

Should this be?
```
  readonly minimum: number | undefined;
  readonly maximum: number | undefined;
```

amlehman: I would just leave them as optional props personally if they can be undefined but that works too.

paribartandhakal: Why I am choosing the latter is because we want to pass in max and min values even though they are undefined. Previously, since it was optional we were not passing it leading to this bug.

amlehman: From a technical perspective there's no difference between passing in undefined and not passing in a prop. But if you want to convey that a value needs to be passed in, that works 👍

wteouchicago:
### Issues
1. Something has gone wrong with Age at Index. It should be like what's in develop (0-89).
- I see the json is messed up here. not sure if that's the cause.
- FWIW, days_to_birth is different from develop (but better behaviour since it's following our dictionary/the json more closely). 
- Maybe what we need to do is make sure the json is correct?

Screenshot 2024-09-10 at 1 13 45 PM

wteouchicago:
looks good! 👍

craigrbarnes (approved):

PR #1283, opened 2024-09-11
PEAR-2141 - Error modals instead of notifications for saving set problems

amlehman:
## Description
Adds the remaining error modals instead of notifications. Created a global modal for save set errors since it is now used in quite a few places. 

## Checklist

- [x] Added proper unit tests
      flagged issues

### DIFF
--- a/packages/portal-proto/src/features/layout/UserFlowVariedPages.tsx
+++ b/packages/portal-proto/src/features/layout/UserFlowVariedPages.tsx
@@ -9,9 +9,12 @@ import {
   selectCohortMessage,
   clearCohortMessage,
   useGetBannerNotificationsQuery,
+  selectCurrentModal,
+  Modals,
+  hideModal,
 } from "@gff/core";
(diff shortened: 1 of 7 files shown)


packages/portal-proto/src/features/layout/UserFlowVariedPages.tsx
```diff
@@ -202,6 +208,19 @@ export const UserFlowVariedPages = ({
           >
             {children}
           </main>
+          <Modal
```
paribartandhakal, line 211: Is it here because only analysis pages can lead to this?

paribartandhakal (approved):

wteouchicago:
Looks good! 👍 

One thing I noticed is that for most situations, after you dismiss the error modal, you return to wherever you were when saving, _except_ when it's related to the modal where you can upload/enter genes/ssms. I think this is fine since these errors should be extremely rare. 🤞

PR #1299, opened 2024-09-23
PEAR-935: add featureFlag support

craigrbarnes:
## Description

This adds a feature flag file to support feature flags and defines a feature flag for Single-Cell RNA-seq.

 The feature flag can be turned on by defining it in localStorage or via an environment variable.
Note that the value of the env variable will be evaluated at build time and will not be changeable after the build.

To turn on  Single-Cell RNA-seq: either add: `GDC_PORTAL_DISPLAY_SC_RNA_SEQ_APP: true` to local storage or:
```
export NEXT_PUBLIC_GDC_PORTAL_DISPLAY_SC_RNA_SEQ_APP=true
npm run dev
```

## Checklist

- [ ] Added proper unit tests
      flagged issues

### DIFF
--- a/packages/portal-proto/src/features/featureFlags.ts
+++ b/packages/portal-proto/src/features/featureFlags.ts
@@ -12,8 +12,8 @@ if (typeof window !== "undefined") localStorage = window.localStorage;
  * Feature flag for Single Cell RNA-seq
  */
 export const DISPLAY_SC_RNA_SEQ_APP =
-  localStorage.GDC_PORTAL_DISPLAY_SC_RNA_SEQ_APP ||
-  process.env.NEXT_PUBLIC_GDC_PORTAL_DISPLAY_SC_RNA_SEQ_APP ||
+  localStorage?.GDC_PORTAL_DISPLAY_SC_RNA_SEQ_APP ||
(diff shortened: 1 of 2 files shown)


paribartandhakal (approved):

packages/portal-proto/src/features/featureFlags.ts
```diff
@@ -0,0 +1,37 @@
...
+let localStorage: Storage = {} as Storage;
+if (typeof window !== "undefined") localStorage = window.localStorage;
+
+/**
+ * Feature flag for Single Cell RNA-seq
+ */
+export const DISPLAY_SC_RNA_SEQ_APP =
+  localStorage.GDC_PORTAL_DISPLAY_SC_RNA_SEQ_APP ||
+  process.env.NEXT_PUBLIC_GDC_PORTAL_DISPLAY_SC_RNA_SEQ_APP ||
+  false;
```
philtom-ctds: Would optional chaining be appropriate here?

```suggestion
export const DISPLAY_SC_RNA_SEQ_APP =
  localStorage?.GDC_PORTAL_DISPLAY_SC_RNA_SEQ_APP ||
  process?.env?.NEXT_PUBLIC_GDC_PORTAL_DISPLAY_SC_RNA_SEQ_APP ||
  false;
```

craigrbarnes: I agree that would be safer.

PR #1304, opened 2024-09-25
Qa 2290 add cart page automation ids

bpierce01:
## Description
- Added IDs to cart page
- Added IDs to header

### DIFF
--- a/packages/portal-proto/src/features/cart/CartHeader.tsx
+++ b/packages/portal-proto/src/features/cart/CartHeader.tsx
@@ -132,6 +132,7 @@ const CartHeader: React.FC<CartHeaderProps> = ({
         <Menu width="target">
           <Menu.Target>
             <Button
+              data-testid="button-download-cart"
               classNames={{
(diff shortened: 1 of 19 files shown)


packages/portal-proto/src/features/cart/CartHeader.tsx
```diff
@@ -476,7 +479,7 @@ const CartHeader: React.FC<CartHeaderProps> = ({
               Remove From Cart
             </Button>
           </Menu.Target>
-          <Menu.Dropdown>
+          <Menu.Dropdown data-testid="dropdown-menu-options">
```
paribartandhakal, line 482: do you wanna differentiate this from the dropdown above?

bpierce01: Nope, only one of those dropdowns can appear at a time. I label every dropdown with the same data-testid so one function can work for all of them.

paribartandhakal (approved):

PR #1309, opened 2024-09-30
PEAR-2186: Update download icon.

paribartandhakal:
## Checklist
      flagged issues

### DIFF
--- a/packages/portal-proto/src/components/DownloadButtons/DownloadButton.tsx
+++ b/packages/portal-proto/src/components/DownloadButtons/DownloadButton.tsx
@@ -1,5 +1,5 @@
 import { ButtonProps, Loader, Tooltip } from "@mantine/core";
-import { FiDownload } from "react-icons/fi";
+import { MdDownload as DownloadIcon } from "react-icons/md";
 import download from "src/utils/download";
 import { hideModal, Modals, useCoreDispatch } from "@gff/core";
 import { Dispatch, SetStateAction, forwardRef } from "react";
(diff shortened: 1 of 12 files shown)


wteouchicago:
looks good -- I didn't see any instances of download icons being unchanged. 👍 

one thing, because the updated icon takes up less width than the old one (at least to my eyes), I wonder if we need to move the icon nearer to the text in buttons. here's an example (Repository):
Screenshot 2024-09-30 at 12 40 01 PM

maybe @cemile-ctds has thoughts.

amlehman:
You don't have to address this here but looking at this PR made me think maybe we should have a file we import our "canonical" versions of the icons we are using all over the app (download, save, close, search, etc.) and then all consuming components could import from there.

amlehman (approved):

paribartandhakal:
> You don't have to address this here but looking at this PR made me think maybe we should have a file we import our "canonical" versions of the icons we are using all over the app (download, save, close, search, etc.) and then all consuming components could import from there.

Good idea and I feel like there's already a ticket for it, right @wteouchicago?

wteouchicago:
> > You don't have to address this here but looking at this PR made me think maybe we should have a file we import our "canonical" versions of the icons we are using all over the app (download, save, close, search, etc.) and then all consuming components could import from there.
> 
> Good idea and I feel like there's already a ticket for it, right @wteouchicago?

I'm not that prescient! But yes, I'll make one ;)

paribartandhakal:
> > > You don't have to address this here but looking at this PR made me think maybe we should have a file we import our "canonical" versions of the icons we are using all over the app (download, save, close, search, etc.) and then all consuming components could import from there.
> > 
> > 
> > Good idea and I feel like there's already a ticket for it, right @wteouchicago?
> 
> I'm not that prescient! But yes, I'll make one ;)

I must be hallucinating. I was so confident I saw it. Maybe it was this ticket.

PR #1312, opened 2024-09-30
PEAR-2022 - Investigate further state-clearing improvements

amlehman:
## Description
The "Unexpected Error” state-clearing escape hatch was only clearing the core persisted store and not local app persisted stores. This is a little hard to test, you can try changing the shape of the store from the slices and then refreshing. 

## Checklist

- [x] Added proper unit tests
      flagged issues

### DIFF
--- a/packages/portal-proto/src/components/ClearStoreErrorBoundary.tsx
+++ b/packages/portal-proto/src/components/ClearStoreErrorBoundary.tsx
@@ -1,7 +1,9 @@
 import React, { Component, ReactNode } from "react";
 import { NextRouter, withRouter } from "next/router";
 import { Modal } from "@mantine/core";
-import { persistor } from "@gff/core";
+import { persistor as corePersistor, persistor } from "@gff/core";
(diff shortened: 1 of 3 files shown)


wteouchicago:
works for me! 👍 I ended up changing the names of various keys in `persist:DownloadApp` in local storage to get the crash I needed.

paribartandhakal (approved):

packages/portal-proto/src/components/ClearStoreErrorBoundary.tsx
```diff
@@ -45,7 +47,14 @@ class ClearStoreErrorBoundary extends Component<
             <ModalButtonContainer data-testid="modal-button-container">
               <DarkFunctionButton
                 onClick={async () => {
-                  await persistor.purge();
+                  if (this.props.router?.query?.app === "Downloads") {
```
craigrbarnes, line 50: The solution is currently adequate since only a few apps have their own data storage. However, as the number of apps increases, we may need to revisit this.

craigrbarnes (approved):

PR #1320, opened 2024-10-02
PEAR-2201 - Remove duplicated data-testid

amlehman:
## Description
Position elements with flex CSS instead of double rendering to fix duplicate test id issue

## Checklist

- [ ] Added proper unit tests
      flagged issues

### DIFF
--- a/packages/portal-proto/src/components/Table/TableHeader.tsx
+++ b/packages/portal-proto/src/components/Table/TableHeader.tsx
@@ -26,6 +26,7 @@ interface TableHeaderProps<TData> {
   columnOrder: string[];
   setColumnOrder: (order: string[]) => void;
   baseZIndex?: number;
+  customBreakpoint?: number;
 }
 
 const TitleWrapper: React.FC<{ title: React.ReactNode }> = ({ title }) => (
@@ -135,6 +136,7 @@ function TableHeader<TData>({
   columnOrder,
   setColumnOrder,
   baseZIndex,
+  customBreakpoint,
 }: TableHeaderProps<TData>) {
   const [searchTerm, setSearchTerm] = useState(search?.defaultSearchTerm ?? "");
   const [searchFocused, setSearchFocused] = useState(false);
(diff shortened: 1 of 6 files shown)


wteouchicago:
Verified I only saw one `text-total-item-count` in the cases table.

I also looked around a bit at various tables (esp their headers). In Repository, this was happening ~1424 width, which I don't think was intended (search bar being on the left). @paribartandhakal would know better. I don't think it's a big deal since it stops happening ~1419, but just in case it reveals something worse. :) 
(oddly it doesn't happen for me if I have devtools open) 
Screenshot 2024-10-02 at 5 47 47 PM

Ditto expanded Cohort Bar > Cases table ~1280.
Screenshot 2024-10-02 at 6 00 12 PM

paribartandhakal:
Screenshot 2024-10-03 at 11 00 05 AM
Maybe we need gap here?

@wteouchicago I am not able to reproduce at 1424px
Screenshot 2024-10-03 at 11 01 44 AM

paribartandhakal:
I verified a few tables and it looks similar to what we had before.

Just to double check we might need @cemile-ctds' eye here. But overall it looks great to me!
Much cleaner than how I approached this problem. 👏

amlehman:
I'm not able to recreate that either @wteouchicago. Weird 🤔

wteouchicago:
> I'm not able to recreate that either @wteouchicago. Weird 🤔

let's blame it on my computer/browser and call it :)

craigrbarnes:
I can reproduce this on the Files Table in the repository.

image

wteouchicago:
FWIW, the weird behaviour happens for me in Chrome, but not in Safari.

cemile-ctds:
The total stats on the file table are longer than those of the others. I believe that’s why it moves up before the breakpoint, and when the breakpoint hits, the gap appears. Can we customize the breakpoint for this table?

P.S. I cannot recreate Wendy and Craig's issue with any of the browsers.

wteouchicago:
It worked! I can no longer get the search bar to show up on the left! 🥳

cemile-ctds:
Looks good on my side.

craigrbarnes (approved):

craigrbarnes (changes requested):
I can still reproduce the issue. Since no one else can, we could merge this and add it as a bug. This PR does fix the original issue.

craigrbarnes (approved):
Approving and will file another ticket for the search bar jump

wteouchicago:
I will file a follow up. The issue is happening for me again (at around 1430 width) - possibly I was adjusting the size of my window too quickly earlier to catch it.

PR #1337, opened 2024-10-09
PEAR-1203: Include timestamp for file manifests

paribartandhakal:
## Checklist
- [X] Added proper unit tests
      flagged issues

### DIFF
--- a/packages/portal-proto/src/utils/index.ts
+++ b/packages/portal-proto/src/utils/index.ts
@@ -289,3 +289,18 @@ export const XL_BREAKPOINT = parseInt(tailwindConfig.theme.extend.screens.xl);
 export const REPO_BREAKPOINT = parseInt(
   tailwindConfig.theme.extend.screens["Custom-Repo-Width"],
 );
+
+// https://stackoverflow.com/questions/2388115/get-locale-short-date-format-using-javascript
+/**
+ * Returns the current local time as a formatted timestamp string.
+ *
+ * @returns A timestamp in the format 'YYYY-MM-DD.HHMMSS'.
+ *
+ * Uses the Swedish ("sv-SE") locale to get a base format of "YYYY-MM-DD HH:MM:SS".
+ */
+export function getFormattedTimestamp() {
+  return new Date()
+    .toLocaleString("sv-SE")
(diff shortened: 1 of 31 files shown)


packages/portal-proto/src/utils/index.ts
```diff
@@ -289,3 +289,18 @@ export const XL_BREAKPOINT = parseInt(tailwindConfig.theme.extend.screens.xl);
...
+
+// https://stackoverflow.com/questions/2388115/get-locale-short-date-format-using-javascript
+/**
+ * Returns the current local time as a formatted timestamp string.
+ *
+ * @returns A timestamp in the format 'YYYY-MM-DD.HHMMSS'.
+ *
+ * Uses the Swedish ("sv-SE") locale to get a base format of "YYYY-MM-DD HH:MM:SS".
+ */
+export function getFormattedTimestamp() {
```
amlehman: We also have convertDateToString to create timestamps: https://github.com/NCI-GDC/gdc-frontend-framework/blob/develop/packages/portal-proto/src/utils/date.ts

I like your implementation better but we should probably only have one timestamp util 😄

amlehman (dismissed):

packages/portal-proto/src/utils/index.ts
wteouchicago: qqn -- have we been using UTC date all over (instead of local date)?

paribartandhakal: We were just using YYYY-MM-DD before but it was UTC I believe.
But because it was just YYYY-MM-DD it didn't matter much IMO. 
Now we are explicitly using local timezone as we require HHMMSS for manifest downloads.

paribartandhakal: @amlehman I update the current file to just use one.

amlehman: Looks great 👍

amlehman (approved):

wteouchicago:
looks good! I tested the manifest downloads and spot checked other downloads. 👍

craigrbarnes (approved):
Code changes are fine

PR #1347, opened 2024-10-14
PEAR-452: Add thousands comma separator for cDAVE and elsewhere

paribartandhakal:
## Checklist
- [ ] Added proper unit tests
      flagged issues


wteouchicago:
1. in cDAVE, the TSV download from the table includes commas for the property values. Can we remove those commas from the TSV download unless it's a pain to do so? While it's not the best TSV for sorting, some people may want to use scripts to parse (and we'll be consistent with other TSV counts). 

2. in cDAVE, the custom bins, we don't need to add the commas for the field inputs (we only want the comma for the "instructions"/information section.). When I type in various numbers like 11111 or 2222.22, I'm getting NaNs in the error messages.

3. I'm having trouble typing in numbers in Cohort Builder and getting it to "accept" the number I type. For example, for Demographic > Age At Diagnosis, when I type in "11111" in the To:, it's becoming "11". Just in case this is due to comma separating -- we don't need to comma separate inputs.

4. Badge number on the Cart in the header doesn't seem to have the comma separator.

paribartandhakal:
CI pipeline failure notice: Need to update typescript after I update mantine. [WIP]

wteouchicago:
> 1. in cDAVE, the TSV download from the table includes commas for the property values. Can we remove those commas from the TSV download unless it's a pain to do so? While it's not the best TSV for sorting, some people may want to use scripts to parse (and we'll be consistent with other TSV counts).

We decided to skip fixing this since the easy way would affect enums that should have commas in them.

Issues 2 - 4 all look good. 👍

craigrbarnes (approved):

PR #1359, opened 2024-10-25
PEAR-2247 - Cohort Builder, endless spinners in some custom filters cards

amlehman:
## Checklist

- [ ] Added proper unit tests
      flagged issues

### DIFF
--- a/packages/core/src/features/facets/data/facet_additional_data.json
+++ b/packages/core/src/features/facets/data/facet_additional_data.json
@@ -467,8 +467,6 @@
     "facet_filter": "family_histories.relationship_age_at_diagnosis",
     "facet_type": "age",
     "label": "Relationship Age at Diagnosis",
-    "minimum": 0,
-    "maximum": 89,
     "property_type": "number"
   },
   "family_histories.relatives_with_cancer_history_count": {
--- a/packages/core/src/features/facets/facetDictionaryApi.ts
+++ b/packages/core/src/features/facets/facetDictionaryApi.ts
@@ -74,6 +74,8 @@ export const classifyFacetDatatype = (f: FacetDefinition): FacetTypes => {
   if (fieldName.includes("is_cancer_gene_census")) return "toggle";
   if (fieldName.includes("figo")) return "enum";
   if (fieldName.includes("age_is_")) return "enum";
+  if (fieldName.includes("age_range")) return "enum";
+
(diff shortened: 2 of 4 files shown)


packages/core/src/features/facets/data/facet_additional_data.json
```diff
@@ -467,8 +467,6 @@
     "facet_filter": "family_histories.relationship_age_at_diagnosis",
     "facet_type": "age",
     "label": "Relationship Age at Diagnosis",
-    "minimum": 0,
-    "maximum": 89,
```
wteouchicago, line 471: is this no longer needed? what happens when we start getting data for this property? will it display a number range card with min 0 and max 89?

wteouchicago:
Verified there aren't endless spinners for:
- `family_histories.relationship_age_at_diagnosis`
- `follow_ups.other_clinical_attributes.undescended_testis_corrected_age_range`
- `diagnoses.pathology_details.epithelioid_cell_percent_range`

Just have a qn about the min and max removal.

packages/core/src/features/facets/data/facet_additional_data.json
amlehman: Yes, it will. The default when we don't have this defined is min 0 and max 89.  There's some logic in the date facets that if the min and max are defined then it converts it to years from days ([here](https://github.com/NCI-GDC/gdc-frontend-framework/blob/PEAR-2247-endless-spinner/packages/portal-proto/src/features/facets/NumericRangeFacet/Years.tsx#L160)). When these are already defined as 0 and 89, it converts that to 0 and 0 in years and then tells the request that it wants to create a range with 0 buckets resulting in this request error: `'No [ranges] specified for the [family_histories.relationship_age_at_diagnosis:range] aggregation`

I tried tweaking that date logic but noticed that incorrectly changed some of the ranges on the other date cards. I think this needs to be fixed for real with better type definitions for these fields so that we don't have this patchy date logic all over that (wrongly sometimes) tries to infer the day/year typing of these fields.

wteouchicago: I'll make a note of a better fix in the spike about improving this area. thanks!

wteouchicago:
All good here. made a note in jira about better fix for all the date logic. 👍

craigrbarnes (approved):
Changes are fine

PR #1364, opened 2024-11-01
bugfix: node-gyp not picking up proxy vars

a-collet:
## Description
node-gyp does not work with proxy vars and always wants to download from nodejs.org. See node-gyp/issues/1133 and 1154

Value of `NODE_VER` is the one that node-gyp tries to pull by default.
## Checklist

- [ ] Added proper unit tests
      flagged issues


craigrbarnes (commented):
Approved

PR #1380, opened 2024-11-26
PEAR-2017 - adds content security policy header

amlehman:
## Checklist

- [ ] Added proper unit tests
      flagged issues

### DIFF
--- a/packages/portal-proto/next.config.js
+++ b/packages/portal-proto/next.config.js
@@ -18,6 +18,21 @@ const buildHash = () => {
   }
 };
 
+const cspHeader = `
+    default-src 'self';
+    script-src 'self' 'unsafe-eval' 'unsafe-inline' https://assets.adobedtm.com https://dap.digitalgov.gov https://www.googletagmanager.com;
+    style-src 'self' 'unsafe-inline';
+    connect-src 'self' https://portal.gdc.cancer.gov https://rum.browser-intake-datadoghq.com https://api.gdc.cancer.gov https://www.google-analytics.com;
+    frame-src https://portal.gdc.cancer.gov;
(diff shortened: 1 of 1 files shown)


packages/portal-proto/next.config.js
```diff
@@ -18,6 +18,21 @@ const buildHash = () => {
   }
 };
 
+const cspHeader = `
+    default-src 'self';
+    script-src 'self' 'unsafe-eval' 'unsafe-inline' https://assets.adobedtm.com https://dap.digitalgov.gov https://www.googletagmanager.com;
+    style-src 'self' 'unsafe-inline';
+    connect-src 'self' https://portal.gdc.cancer.gov https://rum.browser-intake-datadoghq.com https://api.gdc.cancer.gov https://www.google-analytics.com;
+    frame-src https://portal.gdc.cancer.gov;
```
amlehman, line 26: I explicitly declared portal.gdc.cancer.gov to keep some things on dev sites from breaking. We could potentially not add the header on dev sites as well. I'm not crazy about that approach because I think there's a potential for problems to appear on live with the CSP header that could have been caught in dev but I'm curious on other people's opinions.

craigrbarnes: Since @philtom-ctds knows the architecture of GDC the best, he should be able to comment.

philtom-ctds: I think this is ok.  
1. We might see issues if we ever revive the uat subdomains, but we can deal with that when it happens, if ever.
2. Since our QA environments use the same domains as production, we should be able to catch issues in qa that didn't manifest in dev.

philtom-ctds (approved):

PR #1396, opened 2024-12-13
PEAR-2274 -Cohort Builder search - enums 

amlehman:
## Description
Fetch all enums up front for search input to correctly populate search dict

## Checklist

- [ ] Added proper unit tests
      flagged issues

### DIFF
--- a/packages/portal-proto/src/features/cohortBuilder/dictionary.tsx
+++ b/packages/portal-proto/src/features/cohortBuilder/dictionary.tsx
@@ -110,7 +110,11 @@ export const useFacetTabLoaded = (tab: string) => {
     useCoreSelector((state) => selectFacetDefinition(state)).data || {};
   const facetResults = useCoreSelector((state) => selectCaseFacets(state));
   const facetTab = tabsConfig[tab];
-  const facetList = getFacetInfo(facetTab?.facets, {
+  if (facetTab?.facets === undefined) {
+    return false;
+  }
+
+  const facetList = getFacetInfo(facetTab.facets, {
     ...facets,
   });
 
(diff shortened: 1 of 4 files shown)


wteouchicago:
Definitely finding the expected enums that I couldn't find before! 

Getting a crash when I do the following:
- use standard unsaved cohort, no filters
- go to Cohort Builder
- close Cohort Builder
- delete the current cohort
- crash / Unexpected Error 
Screenshot 2024-12-13 at 4 46 18 PM

Also crashes when I do the following (using localhost that allows auth):
- Brain cohort from home page; save
- Eye cohort from home page; save
- While on Eye, go to Cohort Builder
- crash / Unexpected Error
- click Ok on the escape hatch
- on Eye, go to Cohort Builder
- crash / Unexpected Error

amlehman:
Hi @wteouchicago, weirdly I wasn't able to replicate from your steps 🤔 I did make a change that I think should fix it based on that stack trace if you could test when you get a chance.

wteouchicago:
🤔 I'm getting a crash just launching Cohort Builder from the card (not from the header). Not sure if this was happening before. I'm also not too sure which link I used to get to the Cohort Builder when I tested previously. 
Screenshot 2024-12-17 at 1 04 41 PM

If I just stick to the link in the header, I have no problems even when following the previous testing steps.

wteouchicago:
No more crashes - yay!

So...I noticed that the enums on the cards in the Available Data category are not being found when searching, even when I am in the Available Data category.  Is there anything we can do about that?  They are a bit different and are files filters rather than cases filters.

(Arriba is in the Workflow Type card)
Screenshot 2024-12-19 at 11 19 17 AM

amlehman:
Hi @wteouchicago, good catch. I fixed the file type facets.

wteouchicago:
works great! 👍 🔍 👍

packages/portal-proto/src/features/cohortBuilder/dictionary.tsx
```diff
@@ -1,4 +1,4 @@
-import { useMemo } from "react";
+import { useDeepCompareEffect } from "use-deep-compare";
```
paribartandhakal, line 1: Nitpick: didn't notice any JSX in this file. Might as well name `dictionary.tsx`.

paribartandhakal (approved):

PR #1403, opened 2024-12-18
PEAR-2107 - Cohort Builder Search Improvement

amlehman:
## Checklist

- [ ] Added proper unit tests
      flagged issues

https://github.com/user-attachments/assets/83400507-375b-4d1c-a74a-773e6528dedb


wteouchicago:
Looking good!

1. One thing, I noticed that it's auto-searching in the card even when there are no enum results. For example:
- type "as" in the Cohort Builder search
- select the 5th result, which should be Gender. Note that it doesn't find any "as" in the enums/values 
- it goes to Gender with "as" filled in
- what's expected is that it just goes to Gender without filling in "as"

2. _Sometimes_ the page doesn't scroll so that the card is at the top, e.g.
- search for "as"
- go to 4th result -- Site of Resection or Biopsy. It didn't try to scroll for me so that the card is the nearer the top
- go back to the Cohort Builder search and reselect Site of Resection or Biopsy from the results. Now it scrolls so that the card is nearer the top
- I thought at first it would only scroll if I'm already in the category, but then it does scroll if I search for "high risk" while in the General category and select any of the options.
- browser window was at 1310x1044

I didn't personally find the highlight eye-catching enough compared to the mockup (maybe the highlight just needs to be slightly thicker?), but I'll leave this to @cemile-ctds to decide. Also having @cemile-ctds check the screen reader accessibility portion.

cc @ebusygina

wteouchicago:
1. looks good!
2. It's still happening for me, but I'm fine not fixing it if it's a pain -- at least the card is scrolling to somewhere in view.


https://github.com/user-attachments/assets/23ab6050-697a-47ab-94ec-ad14544d98f1



Making the highlight thicker did help in my case, but I'll let @cemile-ctds be the final judge.

amlehman:
Hey @wteouchicago, when you are coming from another tab, all of the facet cards are loading and still determining their height so the scroll doesn't land 100% on the element because other elements are changing the height of the page. I can play around with it a bit more to see if waiting for the cards to load to scroll works better.

wteouchicago:
> Hey @wteouchicago, when you are coming from another tab, all of the facet cards are loading and still determining their height so the scroll doesn't land 100% on the element because other elements are changing the height of the page. I can play around with it a bit more to see if waiting for the cards to load to scroll works better.

Got it - that makes sense. Feel free to play a bit more, but no need to go crazy. :)

cemile-ctds:
1. If there is no matching enum/value in the clicked search result, do we not highlight the card?
2. I also noticed that once the card is highlighted, if I search again and select the same search result, the card doesn't highlight.
3. Lastly, scrolling reduces the visibility of the highlight. Maybe we should increase the duration.

wteouchicago:
> If there is no matching enum/value in the clicked search result, do we not highlight the card?

I think we should highlight in this situation.

> I also noticed that once the card is highlighted, if I search again and select the same search result, the card doesn't highlight.

Does seem like we should highlight (again).

> Lastly, scrolling reduces the visibility of the highlight. Maybe we should increase the duration.

Makes sense to me. Not sure what would work - 7s? 10s?

cemile-ctds:
Looks good on my side.

packages/portal-proto/src/features/cohortBuilder/FacetTabs.tsx
```diff
@@ -298,6 +311,12 @@ export const FacetTabs = (): JSX.Element => {
 
   return (
     <div className="w-100">
+      <span
+        id={`facetTab-liveRegion`}
```
paribartandhakal: Nitpick again 😄 
```suggestion
        id="facetTab-liveRegion"
```

paribartandhakal (approved):
👏

PR #1417, opened 2025-01-07
PEAR-2294 - Quick Search - Randomly, search navigation will not work

amlehman:
## Description
This was caused by some sort of race condition where the file history call was getting fetched/used when it shouldn't be. Wasn't able to pinpoint why that was happening but this at least prevents the error caused by the onclick event using file history data incorrectly in that case. 

## Checklist

- [ ] Added proper unit tests
      flagged issues

### DIFF
--- a/packages/portal-proto/src/components/QuickSearch/QuickSearch.tsx
+++ b/packages/portal-proto/src/components/QuickSearch/QuickSearch.tsx
@@ -205,9 +205,11 @@ export const QuickSearch = (): JSX.Element => {
     if (!id) {
       return;
     }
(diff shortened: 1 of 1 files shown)


wteouchicago:
I was not able to make the error happen, but if it's a race condition issue, I'm not confident it wasn't through good (bad?) luck that it didn't happen to me here. 👍 

If it is still happening though, I'm guessing Brent would be able to catch it in QA with automation., so we should merge and let Brent test if this passes code review.

paribartandhakal (approved):

PR #1418, opened 2025-01-09
GFF-2254: update redux-toolkit to version 2.5.0

craigrbarnes:
## Description
* updates redux-toolkit to 2.5.0
## Checklist

- [ ] Added proper unit tests
      flagged issues

### DIFF
--- a/packages/core/src/listeners.ts
+++ b/packages/core/src/listeners.ts
@@ -1,4 +1,8 @@
-import { AnyAction, createListenerMiddleware, isAnyOf } from "@reduxjs/toolkit";
+import {
+  createListenerMiddleware,
+  isAnyOf,
+  type PayloadAction,
+} from "@reduxjs/toolkit";
 import type { TypedStartListening } from "@reduxjs/toolkit";
 import Cookies from "js-cookie";
 import { CoreDispatch } from "./store";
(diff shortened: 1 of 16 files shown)


packages/core/src/listeners.ts
```diff
@@ -103,8 +103,9 @@ startCoreListening({
   ),
   effect: async (action, listenerApi) => {
     // the last cohort added is the one we want to get the case count for
-
-    const cohortId = action?.payload?.id;
+    console.log("startCoreListening", action);
```
paribartandhakal: log alert!

paribartandhakal:
```suggestion
```

packages/portal-proto/package.json
```diff
@@ -30,7 +30,7 @@
     "@mantine/hooks": "^7.14.3",
     "@mantine/notifications": "^7.14.3",
     "@react-spring/web": "^9.5.5",
-    "@reduxjs/toolkit": "^1.8.5",
+    "@reduxjs/toolkit": "^2.5.0",
```
paribartandhakal, line 33: just wanted to understand why we need it in both the places (other being at core's `package.json`).

craigrbarnes: There are a couple of locations in portal-proto where we need the types defined in redux-toolkit.
for example UpdateFiltersButton

packages/core/src/listeners.ts
craigrbarnes: Removed. It reminded me to refactor that listener

paribartandhakal (approved):

PR #1424, opened 2025-01-16
PEAR-2293: Fix for Cohort Bar: Unique Name Replacement behaves incorrectly under certain steps 

paribartandhakal:
## Description
**Request for while testing:**
Test the scenarios of add, save, save as, replacement and deletion. 🙏 

## Checklist
- [ ] Added proper unit tests
      flagged issues


wteouchicago:
Verified the bug Brent described no longer happens.

I also tried various saves, save-as, replaces, add tos and remove froms and they worked ok.

Looks good! 👍

packages/portal-proto/src/components/Modals/SaveCohortModal/types.ts
```diff
@@ -0,0 +1,20 @@
+import { FilterSet } from "@gff/core";
+
+export interface SaveCohortModalProps {
+  opened: boolean;
```
amlehman: This is a subjective style thing but personally I don't like moving the prop definition to a different file. I would rather not have to flip to a different file to see the props for a component and I think its more likely that the props and jsdoc comment drift out of sync.

packages/portal-proto/src/components/Modals/SaveCohortModal/index.tsx
```diff
@@ -0,0 +1,241 @@
...
+    fetchSavedFilters,
+    coreDispatch,
+  } = useCohortOperations();
+  const [addCohort, { isLoading: isAddCohortLoading }] = useAddCohortMutation();
+  const [
+    fetchCohortList,
+    { isSuccess: isCohortListSuccess, isLoading: isCohortListLoading },
+  ] = useLazyGetCohortsByContextIdQuery();
+
+  const [state, setState] = useState<CohortState>({
```
amlehman, line 79: Why put all of these into one piece of state? Their states are not related most of the time.

packages/portal-proto/src/components/Modals/SaveCohortModal/index.tsx
```diff
@@ -0,0 +1,241 @@
...
+    state.cohortSavedMessage,
+  ]);
+
+  const handleSaveAction = async (newName: string, replace: boolean) => {
+    try {
+      let cohortFilters = filters;
+
+      if (createStaticCohort) {
+        try {
+          cohortFilters = await createStaticCohortSet(filters, caseFilters);
```
amlehman, line 140: I like extracting out and naming these, makes this piece of code easier to follow 👍

paribartandhakal: I found it a bit easier to keep track of them. We are updating them together at many instances.

packages/portal-proto/src/components/Modals/SaveCohortModal/index.tsx
paribartandhakal: I found it a bit easier to keep track of them. We are updating them together at many instances.

packages/portal-proto/src/components/Modals/SaveCohortModal/types.ts
paribartandhakal: I agree. I am bringing it back to the same file.
I thinking making things clear got in my head 😄 .

craigrbarnes: I agree. I like the component props to be close in code to the component, especially since, in this case, the props type is not used anywhere else.  Depending on what you do with the state, I would do the same for CohortState

craigrbarnes (approved):
Code changes and refactoring are appropriate.

PR #1452, opened 2025-02-17
PEAR-2237 - Extract Cohort Workflow into a Shared Component

amlehman:
## Description
- Moves components for managing cohorts to portal-components
- Creates hooks in portal-proto for interacting with the store/backend
- Mocks out a simple cohort workflow in enclave portal
- Removes cohort notification/error handling from the store and into the shared components

## Checklist

- [ ] Added proper unit tests
      flagged issues

Screenshot 2025-02-17 at 10 35 20 AM


wteouchicago:
Some observations -- not sure how much is supposed to be working by now, so there's no need to fix them now if things are not meant to be working yet.

1. Cohort deletion is not always successful. Steps:
- A: start with all the mock cohorts
- B: while on Baily's Cohort, delete the cohort. this seems successful.
- C: while on Pancreas, delete the cohort. The notification says it's deleted by the cohort is still in the bar.
- D: delete Pancreas again. Now it does seem to be deleted.

If at step C, instead of deleting Pancreas right away, you switch to Lung and then delete Lung, then Lung will be successfully deleted.

2. Save As isn't quite working. I'm guessing the actual saving isn't meant to be working yet, but here's an oddity:
- while on Baily's Cohort, do  Save As. name it something else, e.g. "foo", before saving
- it tells me I'm replacing an existing cohort

3. Maybe related to 2. 
- switch to Pancreas (it's unsaved)
- try to save and it'll tell you you are replacing an existing cohort (even if you change the name)

____
I did do a quick run through v2's cohort workflow and things seem to still be working as expected.

amlehman:
Hi @wteouchicago. Yeah, the cohort workflow on the enclave side is just meant to be slightly mocked out, not fully implemented. The save as weirdness is me implementing the mock save as function as always returning that the cohort should be replaced 😄

packages/portal-components/src/cohort/CustomCohortSelectItem.unit.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  it("Unsaved Icon should NOT be visible when the cohort has NOT been modified", () => {
+    const menu_items = {
+      value: "testId",
+      label: "test",
+      isSavedUnchanged: true,
+      cohortStatusMessage: "Changes not saved",
+    };
+    const { queryByTitle, getByText } = render(
+      <MantineProvider>
```
paribartandhakal, line 34: We can move this to a test setup like we have for portal-proto later just so that we don't have to wrap it everywhere.

packages/portal-components/src/common/DropdownMenu/index.ts
```diff
@@ -0,0 +1,3 @@
+import DropdownMenu from "./DropdownMenu";
+
+export default DropdownMenu;
```
paribartandhakal, line 3: should we just rename the main file to be `index.ts`? Not sure if we need this extra file.

packages/portal-components/src/modals/GenericCohortModal.tsx
```diff
@@ -0,0 +1,65 @@
...
+}) => {
+  return (
+    <Modal
+      title={title}
+      opened={opened}
+      padding={0}
+      radius="md"
+      onClose={onClose}
+    >
+      <div className="font-montserrat py-5 pr-6 pl-2.5">
```
paribartandhakal, line 31: I just noticed that we have duplicate fonts in tailwind config. Removing the duplicates might induce other errors so we can do it separately but I wanted to point it out.
```
      fontFamily: {
        montserrat: ["Montserrat", "sans-serif"],
        heading: ["Montserrat", "sans-serif"],
        content: ["Noto Sans", "sans-serif"],
        "content-noto": ["Noto Sans", "sans-serif"],
      },
```

packages/portal-components/src/modals/index.ts
```diff
@@ -0,0 +1,3 @@
+import SaveCohortModal from "./SaveCohortModal";
+
+export { SaveCohortModal };
```
paribartandhakal, line 3: Same comment as above about saving the main file as `index.ts`.

paribartandhakal (approved):
🥇

packages/portal-components/src/common/DropdownMenu/index.ts
amlehman: Personally, I don't like having components in index.ts 🤷‍♀️ One function of this pattern is so we can select what we export. Since this is a library, we don't necessarily want to expose to the consumer everything from the component file (i.e we might need to export a constant or type for a test but it isn't meant to be used outside the library). Also, I think not having a descriptive filename makes files harder to work with in IDEs.

craigrbarnes (approved):
Good job on a tedious task.

PR #1456, opened 2025-02-19
PEAR-2325 - Cohort Builder Search morphology fix

amlehman:
## Description
Changes the the way the search tokens are created by removing most punctuation as a place to split on so that "8010/3" is processed as "8010/3" instead of "8010", "3"

## Checklist

- [x] Added proper unit tests
      flagged issues

Screenshot 2025-02-19 at 1 47 46 PM

### DIFF
--- a/packages/portal-proto/src/features/cohortBuilder/dictionary.tsx
+++ b/packages/portal-proto/src/features/cohortBuilder/dictionary.tsx
@@ -62,6 +62,7 @@ export interface FacetSearchDocument {
 export const miniSearch = new MiniSearch<FacetSearchDocument>({
   fields: ["name", "description", "enum"], // fields to index for full-text search
   storeFields: ["name", "category", "categoryKey", "description", "enum", "id"], // fields to return with search results
+  tokenize: (string) => string.split(/[\n\r\s,]+/u), // indexing tokenizer
   processTerm: (term) => (STOP_WORDS.has(term) ? null : term.toLowerCase()), // index term processing
   searchOptions: {
     boost: {
--- a/packages/portal-proto/src/features/cohortBuilder/dictionary.tsx
+++ b/packages/portal-proto/src/features/cohortBuilder/dictionary.tsx
@@ -50,6 +50,7 @@ export const get_facets = (
 };
 
 export const STOP_WORDS = new Set(["of", "or"]);
+export const TOKENIZE_STRING = /[\n\r\s,\-_]+/u;
 
 export interface FacetSearchDocument {
   name: string;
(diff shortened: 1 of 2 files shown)


wteouchicago:
Changes work well to deal with the weirdness of the enums in Morphology and similar.

Not due to this change, but I'm guessing it was when we switched to minisearch for the card search:
- I noticed the card search no longer finds results if the input is not found at the beginning of the enum , e.g. if I search for "male" in Gender, only "male" is returned but not "female". 
- But searching for "male" in the big search does find "female". 

This feels awkward when I search for "male" in the big search, see that it finds "female", click on the result and see only "male" in the card.

Is there a way to make the cards search more on par with big search in this aspect?

__

Found this; not sure how applicable it still is: https://github.com/lucaong/minisearch/issues/194#issuecomment-1369034130

wteouchicago:
It's working 50000x better now! 👍 

noticed an edge case, no need to fix unless it's super easy: in the card search, it doesn't seem to find results if your input is a number and that number isn't the first letter or after a punctuation. E.g.
- in General Diagnosis > Tumor Grade, search for "1". notice it doesn't find "g1"
- Disease Status and History > Ishak Fibrosis Score, search for "2" and it correctly finds "1,2 - portal fibrosis"

amlehman:
@wteouchicago. As far as I can tell, this intentional behavior by minisearch to not search for characters at the end of the string: https://github.com/lucaong/minisearch/issues/243#issuecomment-1842522975. "x" also won't return a result for gx on Tumor Grade. The second example is working because I have it set up to split the stored terms on a comma (i.e "1,2 - portal fibrosis" will become two tokens "1" "2 - portal fibrosis" and the search is matching the second token where 2 is at the beginning of the term). If we don't want this behavior for some reason, I can try and process the terms differently and use another character to split on.

wteouchicago:
> @wteouchicago. As far as I can tell, this intentional behavior by minisearch to not search for characters at the end of the string: [lucaong/minisearch#243 (comment)](https://github.com/lucaong/minisearch/issues/243#issuecomment-1842522975). "x" also won't return a result for gx on Tumor Grade. The second example is working because I have it set up to split the stored terms on a comma (i.e "1,2 - portal fibrosis" will become two tokens "1" "2 - portal fibrosis" and the search is matching the second token where 2 is at the beginning of the term). If we don't want this behavior for some reason, I can try and process the terms differently and use another character to split on.

Ah! What we have is already very good, I think. We can tweak later if we get feedback. Thanks!

craigrbarnes:
Note: search is a bit more restricted than before. For example, a search for EAGLE no longer returns Program,  Project

wteouchicago:
> Note: search is a bit more restricted than before. For example, a search for EAGLE no longer returns Program, Project

Nice catch! we should try to fix this (in big search and enum search). I noticed "LUAD" isn't finding "APOLLO-LUAD", etc. do we just need to split on "-" and "_" as well?

wteouchicago:
works mo betta again! 👍 👍 🔍 👍 👍

craigrbarnes (approved):
The changes look fine. 
The EAGLE has landed

PR #1484, opened 2025-03-07
PEAR-2101 + 2100: Responsive Cohort Comparison

paribartandhakal:
## Checklist
- [ ] Added proper unit tests
      flagged issues


cemile-ctds (approved):

wteouchicago:
looks good. one small thing: when I mouse over a "0" in the Venn diagram (try demo mode for Cohort Comparison), the tooltip kind of...flies in. any idea what's causing it?

paribartandhakal:
> looks good. one small thing: when I mouse over a "0" in the Venn diagram (try demo mode for Cohort Comparison), the tooltip kind of...flies in. any idea what's causing it?

I noticed this in prod too. I am not quite sure what's causing it. I will try to see.

paribartandhakal:
> > looks good. one small thing: when I mouse over a "0" in the Venn diagram (try demo mode for Cohort Comparison), the tooltip kind of...flies in. any idea what's causing it?
> 
> I noticed this in prod too. I am not quite sure what's causing it. I will try to see.

@wteouchicago This is fixed and much better now. Also, I was wondering should we investigate more (later on) on how to make the venn diagram portion show the tooltip rather than just a value?

wteouchicago:
Looks good! No more flying tooltip! 👍 

> @wteouchicago This is fixed and much better now. Also, I was wondering should we investigate more (later on) on how to make the venn diagram portion show the tooltip rather than just a value?

Hmm, probably unnecessary. I mean, the "0" is the right thing to show for what the Venn diagram is supposed to show.

craigrbarnes (approved):

PR #1490, opened 2025-03-17
PEAR-2266: Fix for sorting by case count issue in cart.

paribartandhakal:
## Checklist
- [ ] Added proper unit tests
      flagged issues

Screenshot 2025-03-18 at 11 38 09 AM

### DIFF
--- a/packages/portal-proto/src/features/cart/ProjectTable.tsx
+++ b/packages/portal-proto/src/features/cart/ProjectTable.tsx
@@ -14,7 +14,7 @@ const ProjectTable: React.FC<ProjectTableProps> = ({
   customDataTestID,
 }: ProjectTableProps) => {
   const sortedProjectData = [...(projectData || [])].sort((a, b) => {
-    return b.file_size - a.file_size;
+    return b.case_count - a.case_count;
   });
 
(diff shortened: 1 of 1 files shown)


packages/portal-proto/src/features/cart/ProjectTable.tsx
```diff
@@ -13,7 +13,11 @@ const ProjectTable: React.FC<ProjectTableProps> = ({
   projectData,
   customDataTestID,
 }: ProjectTableProps) => {
-  const tableData = (projectData || []).map((project) => ({
+  const sortedProjectData = [...projectData].sort((a, b) => {
```
amlehman: Can projectData be undefined here? It looks like the previous way this line was written was guarding against that. You'll get an exception when trying to spread here (and I think a typescript error in strict mode) if it is undefined.

paribartandhakal: Good call. I added the check.

wteouchicago:
we should be sorting by descending number of cases. :P

wteouchicago:
looks good! 👍

amlehman (approved):

craigrbarnes (approved):

PR #1527, opened 2025-05-02
PEAR-2046: Fix for all values are unmatched when passing in over 100k identifiers

paribartandhakal:
## Checklist
- [ ] Added proper unit tests
      flagged issues

### DIFF
--- a/packages/portal-proto/src/components/InputEntityList/InputEntityList.tsx
+++ b/packages/portal-proto/src/components/InputEntityList/InputEntityList.tsx
@@ -199,6 +199,31 @@ const InputEntityList: React.FC<InputEntityListProps> = ({
     [processInputDebounced],
   );
 
+  const handleKeyDown = useCallback(
+    (event: React.KeyboardEvent<HTMLTextAreaElement>) => {
+      if (state.tokens.length >= MATCH_LIMIT) {
+        const allowedKeys = [
+          "Backspace",
+          "Delete",
+          "ArrowLeft",
+          "ArrowRight",
+          "ArrowUp",
+          "ArrowDown",
+          "Home",
+          "End",
+          "Tab",
+        ];
+
+        const isKeyCombo = event.ctrlKey || event.metaKey || event.altKey;
+
+        if (!allowedKeys.includes(event.key) && !isKeyCombo) {
+          event.preventDefault();
+        }
+      }
+    },
+    [],
+  );
+
   const handleFileChange = useCallback(
     async (file: File | null) => {
       dispatch({ type: "SET_FILE", payload: file });
@@ -313,28 +338,7 @@ const InputEntityList: React.FC<InputEntityListProps> = ({
               onChange={(event) => {
                 handleInputChange(event.target.value);
               }}
-              onKeyDown={(event) => {
-                if (state.tokens.length >= MATCH_LIMIT) {
-                  const allowedKeys = [
-                    "Backspace",
-                    "Delete",
-                    "ArrowLeft",
-                    "ArrowRight",
-                    "ArrowUp",
-                    "ArrowDown",
-                    "Home",
-                    "End",
-                    "Tab",
-                  ];
-
-                  const isKeyCombo =
-                    event.ctrlKey || event.metaKey || event.altKey;
-
-                  if (!allowedKeys.includes(event.key) && !isKeyCombo) {
-                    event.preventDefault();
-                  }
-                }
-              }}
+              onKeyDown={handleKeyDown}
               minRows={5}
               maxRows={5}
               id="identifier-input"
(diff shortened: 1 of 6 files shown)


wteouchicago:
1. I found this part a little awkward --  when I reach 50k identifiers and I type some more text, it displays that text then erases it. It's one thing to not allow users to type in more text, but it's another thing to allow them to type something, then automatically erase it. And it doesn't quite match the requirement "Prevent users from typing beyond the 50k limit. Users should not be able to type more than 50k identifiers in the input field.". Is there any adjustment we can make to the behaviour here? We may not have to fix since users who reach 50k ids are likely copy-pasting or uploading all at once.  cc @ebusygina 

Looks good otherwise. I couldn't find a way to trigger the "Error processing file: please check your file and try again.” error, btw.

paribartandhakal:
> 1. I found this part a little awkward --  when I reach 50k identifiers and I type some more text, it displays that text then erases it. It's one thing to not allow users to type in more text, but it's another thing to allow them to type something, then automatically erase it. And it doesn't quite match the requirement "Prevent users from typing beyond the 50k limit. Users should not be able to type more than 50k identifiers in the input field.". Is there any adjustment we can make to the behaviour here? We may not have to fix since users who reach 50k ids are likely copy-pasting or uploading all at once.  cc @ebusygina
> 
> Looks good otherwise. I couldn't find a way to trigger the "Error processing file: please check your file and try again.” error, btw.

I fixed this.

The solution was already in place, but I had overlooked passing the correct dependencies to one of the callbacks.

Also, once we reach 50K identifiers I have allowed the following keys to work:
`["Backspace",
          "Delete",
          "ArrowLeft",
          "ArrowRight",
          "ArrowUp",
          "ArrowDown",
          "Home",
          "End",
          "Tab",
        ];` along with  modifier keys like `Ctrl`, `Cmd` and `Alt`, since users may want to navigate and delete using keyboard.

wteouchicago:
Ah, that works. It may be a bit weird that they can't edit anything (other than delete) once they hit 50k, but that's pretty edge case, plus they can also delete an identifier and add it back with whatever edits.

looks good! 👍

wteouchicago:
Changes made it much nicer. Thanks! 👍

paribartandhakal:
> Changes made it much nicer. Thanks! 👍

Thank you for your idea 😄 !

packages/portal-proto/src/components/InputEntityList/InputEntityList.tsx
```diff
@@ -83,62 +82,180 @@ const InputEntityList: React.FC<InputEntityListProps> = ({
...
       mappedToFields,
       matchAgainstIdentifiers,
       outputField,
       searchField,
-      entityType,
-      setTokens,
     ],
   );
 
+  const procesInput = useCallback(
```
amlehman: "process" is misspelled here.

packages/portal-proto/src/components/InputEntityList/InputEntityList.tsx
```diff
@@ -275,31 +371,40 @@ const InputEntityList: React.FC<InputEntityListProps> = ({
...
-            unmatched={unmatched}
-            numberInput={tokens.length}
-            entityLabel={entityLabel}
-            fieldDisplay={fieldDisplay}
-          />
+          (state.matched.length > 0 || unmatched.length > 0) && (
+            <MatchTablesWrapper
+              matched={state.matched}
+              unmatched={unmatched}
+              numberInput={new Set(state.tokens).size}
```
amlehman, line 396: Why not just use the length of the array here?

packages/portal-proto/src/components/InputEntityList/InputEntityList.tsx
```diff
@@ -83,62 +82,180 @@ const InputEntityList: React.FC<InputEntityListProps> = ({
...
-            entityType,
-            {
-              filters: {
-                op: "in",
-                content: {
-                  field: searchField,
-                  value: uniq(queryTokens.map((t) => t.toLowerCase())),
-                },
+  const processTokens = useCallback(
+    (tokens: string[]): [string[], string | null] => {
```
amlehman, line 86: I like the use of a reducer here 👍 makes it very clear what is going on

packages/portal-proto/src/components/InputEntityList/InputEntityList.tsx
paribartandhakal: Duplicates affect the `numMatched` count in `MatchTablesWrapper`, even though we ignore duplicates during the actual identifier matching, leading to discrepancies.

amlehman (approved):

packages/portal-proto/src/components/InputEntityList/utils.ts
```diff
@@ -152,3 +143,16 @@ const findAllIdentifiers = (
     }
   });
 };
+
```
craigrbarnes, line 146: 👍

craigrbarnes (approved):
Nice solution to this issue.

PR #1532, opened 2025-05-07
Qa 2361 add initial authenticated tests

bpierce01:
## Description
- Added initial controlled-access test

How to Test:
qa-int, vpn2
Near the start of the test, after the cohort is made, it will prompt you to login to the data portal (the login screen will appear). Login with your credentials and then the test will continue uninterrupted. 

gauge run specs/gdc_data_portal_v2/file_summary/download_controlled_access.spec

### DIFF
--- a/holmes-py/README.md
+++ b/holmes-py/README.md
@@ -157,6 +157,26 @@ consistency of test results. Anymore than 6, and there would be some flakey test
 
 If a test does fail, rerun them individually before reporting results.
 
+### How to Execute Each Type of Test Suite
+The tests should be ran by tag. We have open-access and controlled-access tests that are not compatible to be ran together.
+The controlled-access tests require you to manually login when prompted. The other tests do not have to be interacted with
+once execution begins. Any of the commands below can also be ran in parallel or with de-bug mode. See above for commands.
+
+#### Regression Test
+````
+gauge run specs --tags "regression"
+````
+
+#### Data Test
+````
+gauge run specs --tags "data-release"
+````
+
+#### Controlled Access Test
+````
+gauge run specs --tags "controlled-access"
+````
+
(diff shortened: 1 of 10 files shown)


ManjunathJayanna (commented):
During the test, i was able to login and download the controlled file, but do you think there should be a logout of the user at the end of this test or somewhere?

Also, I have not checked. But not sure where the login piece/tests fall when you run the entire suite of tests. 

Because, when you run the entire suite and login happens in the middle somewhere, the rest of the tests will probably/maybe use the same login session to continue with the remaining tests may be, and currently running tests when a user is logged slows down the tests as there is a real performance issue when the user is logged in and browsing becomes very slow.

bpierce01:
> During the test, i was able to login and download the controlled file, but do you think there should be a logout of the user at the end of this test or somewhere?
> 
> Also, I have not checked. But not sure where the login piece/tests fall when you run the entire suite of tests.
> 
> Because, when you run the entire suite and login happens in the middle somewhere, the rest of the tests will probably/maybe use the same login session to continue with the remaining tests may be, and currently running tests when a user is logged slows down the tests as there is a real performance issue when the user is logged in and browsing becomes very slow.

I wrote in the READme how I currently conceive of the tests. For now, the open and controlled access tests are separate and will be ran separately for those reasons you described. We can run regression tag for all open access tests, and run controlled-access tag for these sorts of tests. 

I want to login once per window of tests (you can run in parallel but would have to login each window. I tested to make sure that scenario behaved correctly and it does). Logging in and logging out would lengthen the execution time, and make it much harder on us to execute as we would have to enter our credentials multiple times.

ManjunathJayanna (approved):
Logout will still be good. But approving.

PR #1577, opened 2025-06-24
update the proteinpaint-client version to 2.129.5-c5a0eaea6.1

siosonel:
## Description

Fix the missing "load gene sets" from oncoMatrix gene set edit UI ( https://gdc-ctds.atlassian.net/browse/SV-2620.)

## Checklist

- [ ] Added proper unit tests
      flagged issues

### DIFF
--- a/packages/portal-proto/src/features/proteinpaint/dev.sh
+++ b/packages/portal-proto/src/features/proteinpaint/dev.sh
@@ -7,7 +7,7 @@ if [[ "$1" == "unlink" ]]; then
 	# to test the published client package before submitting a PR with an updated pp-client version
 	npm unlink ../proteinpaint/client
 	npm uninstall @sjcrh/proteinpaint-client --save --workspace=packages/portal-proto
-	npm install @sjcrh/proteinpaint-client@2.129.1 --save --workspace=packages/portal-proto
+	npm install @sjcrh/proteinpaint-client --save --save-exact --workspace=packages/portal-proto
 
 else
 	# to test the local PP client code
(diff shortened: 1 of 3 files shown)


packages/portal-proto/src/features/proteinpaint/dev.sh
```diff
@@ -7,7 +7,7 @@ if [[ "$1" == "unlink" ]]; then
 	# to test the published client package before submitting a PR with an updated pp-client version
 	npm unlink ../proteinpaint/client
 	npm uninstall @sjcrh/proteinpaint-client --save --workspace=packages/portal-proto
-	npm install @sjcrh/proteinpaint-client@2.129.1 --save --workspace=packages/portal-proto
+	npm install @sjcrh/proteinpaint-client --save --save-exact --workspace=packages/portal-proto
```
paribartandhakal, line 10: I am a bit uncertain on whether both `--save` and `--save-exact` is needed here or not.

siosonel: I think I've tried only `--save-exact` by itself a few weeks ago, and the version was still saved as `"^2.129.4"`. I want to pin to an exact version, not a loose match. Per https://docs.npmjs.com/cli/v11/commands/npm-install, it seems `--save-exact` is meant to be used in addition to `---save-prod` (which is the same as `--save`).

```text
When using any of the above options to save dependencies to your package.json, there are two additional, optional flags:

-E, --save-exact: Saved dependencies will be configured with an exact version rather than using npm's default semver range operator.
-B, --save-bundle: Saved dependencies will also be added to your bundleDependencies list.
```

paribartandhakal: Okay, good to know. Just wanted to make sure. Thanks @siosonel

paribartandhakal (approved):

PR #1594, opened 2025-07-24
OncoMatrix and Gene Expression apps should cancel network requests when closed

siosonel:
## Description

Fixes https://gdc-ctds.atlassian.net/browse/SV-2633. 

Tested with:
- open the analysis tools page and the Network tab in Chrome dev tools
- click on the OncoMatrix or Gene Expression card to open the tool
- while the topMutatedGenes or topVariablyExpressedGenes is still loading, close the tool to return to the analysis page: the network request should get canceled and allow other GFF tools to load quickly
- click on the tool card again, but wait after the top genes request is done, but close the tool while the /termdb?for=matrix or /termdb/cluster is still loading: the network request should get canceled and allow other GFF tools to load quickly

## Checklist

- [ ] Added proper unit tests
      flagged issues

### DIFF
--- a/packages/portal-proto/src/features/proteinpaint/MatrixWrapper.tsx
+++ b/packages/portal-proto/src/features/proteinpaint/MatrixWrapper.tsx
@@ -1,4 +1,4 @@
-import { useRef, useCallback, useState, useEffect, FC } from "react";
+import { useRef, useCallback, useState, FC } from "react";
 import { useDeepCompareEffect } from "use-deep-compare";
 import { bindProteinPaint } from "@sjcrh/proteinpaint-client";
 import { useIsDemoApp } from "@/hooks/useIsDemoApp";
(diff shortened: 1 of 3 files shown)


packages/portal-proto/src/features/proteinpaint/MatrixWrapper.tsx
```diff
@@ -218,7 +219,22 @@ export const MatrixWrapper: FC<PpProps> = (props: PpProps) => {
           // in case of race condition
           return prevData.current != data;
         },
-      });
+      })
+        .then?.((_app) => {
+          toolApp.current = _app;
+        })
+        .catch((e) => {
+          throw e;
```
craigrbarnes: I am concerned about throwing an error here, as I am unsure how it will be handled.  We don't throw unhandled exceptions in the portal-prototype code.

siosonel: I'll fix

congyu-lu (approved):

packages/portal-proto/src/features/proteinpaint/MatrixWrapper.tsx
siosonel: converted to `console.error()` instead of throw, see comments in my comment, this catch is not likely to be used except for bundling issues that are not detected during build

craigrbarnes (approved):

PR #1595, opened 2025-07-24
PEAR-2462: Clinical Data Analysis - add option to hide properties with no data

paribartandhakal:
## Checklist
- [ ] Added proper unit tests
      flagged issues


wteouchicago:
**Issues**
1. When a search is applied and a card/category has no results found (whether or not the new checkbox is selected), there is now text in the card that says "No properties with data". This is when there are _some_ results found via the search, but not in every card/category. This isn't quite right. We should either keep the old behaviour (where a card without results found by search would just not be shown -- this is preferred), or we should change the text in this situation to something like "No results found".
2. I think the checkbox label isn't quite the same as the mockup. @cemile-ctds to check
3. Should "No results found" be Montserrat or Noto-Sans? @cemile-ctds

paribartandhakal:
> **Issues**
> 
> 1. When a search is applied and a card/category has no results found (whether or not the new checkbox is selected), there is now text in the card that says "No properties with data". This is when there are _some_ results found via the search, but not in every card/category. This isn't quite right. We should either keep the old behaviour (where a card without results found by search would just not be shown -- this is preferred), or we should change the text in this situation to something like "No results found".
> 2. I think the checkbox label isn't quite the same as the mockup. @cemile-ctds to check
> 3. Should "No results found" be Montserrat or Noto-Sans? @cemile-ctds

I fixed 1 and 2. I will wait for @cemile-ctds' response to fix 3.

cemile-ctds:
The texts _'No results found'_ and _'No properties with data'_ should use the Noto font family.

paribartandhakal:
> The texts _'No results found'_ and _'No properties with data'_ should use the Noto font family.

I update the font.
cc. @wteouchicago

wteouchicago:
Issues 1-3 all look good.

Qn, what size should the "No properties with data" text be? It's a bit bigger than the mockup. Looks ok to me, but I'm not an expert on this. @cemile-ctds ?

Screenshot 2025-08-06 at 5 53 29 PM

cemile-ctds:
> Qn, what size should the "No properties with data" text be? It's a bit bigger than the mockup. Looks ok to me, but I'm not an expert on this. @cemile-ctds ?

The class we should use for the text size is `text-sm`.

amlehman:
@paribartandhakal, can you point out the specific changes for this ticket? It's hard to tell with the moved files.

packages/portal-proto/src/features/cDave/Controls/Controls.tsx
```diff
@@ -0,0 +1,173 @@
...
+const Controls: React.FC<ControlsProps> = ({
+  updateFields,
+  cDaveFields,
+  fieldsWithData,
+  activeFields,
+  controlsExpanded,
+  setControlsExpanded,
+}) => {
+  const [searchTerm, setSearchTerm] = useState("");
+  const [onlyDataChecked, setOnlyDataChecked] = useState(false);
```
paribartandhakal, line 35: added new state var.

packages/portal-proto/src/features/cDave/Controls/Controls.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+  const filteredFields = useDeepCompareMemo(() => {
```
paribartandhakal, line 37: updated this function to return fields based on the added state var and search

packages/portal-proto/src/features/cDave/Controls/Controls.tsx
```diff
@@ -0,0 +1,173 @@
...
+          .includes(searchTerm.toLowerCase()),
+    );
+  }, [searchTerm, cDaveFields, onlyDataChecked, fieldsWithData]);
+
+  const groupedFilteredFields = useDeepCompareMemo(
+    () => groupBy(filteredFields, "field_type"),
+    [filteredFields],
+  );
+
+  const hasSomeResults = Object.values(groupedFilteredFields).some(
```
paribartandhakal, line 61: added this variable to conditionally render `ControlGroup`.

packages/portal-proto/src/features/cDave/Controls/ControlGroup.tsx
```diff
@@ -0,0 +1,169 @@
...
+  readonly name: string;
+  readonly fields: CDaveField[];
+  readonly updateFields: (field: string) => void;
+  readonly activeFields: string[];
+  readonly searchTerm?: string;
+}
+
+const ControlGroup: React.FC<ControlGroupProps> = ({
+  name,
+  fields,
```
paribartandhakal, line 95: now the fields in control group do not need to be further processed like before.

packages/portal-proto/src/features/cDave/Controls/Controls.tsx
```diff
@@ -0,0 +1,173 @@
...
+            )
+          }
+          aria-label="Search fields"
+        />
+
+        <p data-testid="text-fields-with-values" className="py-2 font-heading">
+          <strong>{Object.keys(fieldsWithData).length}</strong> of{" "}
+          <strong>{cDaveFields.length}</strong> properties with data
+        </p>
+        <Checkbox
```
paribartandhakal, line 123: added this

packages/portal-proto/src/features/cDave/Controls/Controls.tsx
```diff
@@ -0,0 +1,173 @@
...
+                <CloseIcon aria-label="clear search" />
+              </ActionIcon>
+            )
+          }
+          aria-label="Search fields"
+        />
+
+        <p data-testid="text-fields-with-values" className="py-2 font-heading">
+          <strong>{Object.keys(fieldsWithData).length}</strong> of{" "}
+          <strong>{cDaveFields.length}</strong> properties with data
```
paribartandhakal, line 121: updated the text here

paribartandhakal:
> @paribartandhakal, can you point out the specific changes for this ticket? It's hard to tell with the moved files.

I have left comments on the changes I made. Not sure why test file is correctly updated to show only the updated changes after I moved it, but not the Control file. This is my nemesis.

paribartandhakal:
> > Qn, what size should the "No properties with data" text be? It's a bit bigger than the mockup. Looks ok to me, but I'm not an expert on this. @cemile-ctds ?
> 
> The class we should use for the text size is `text-sm`.

I have updated the font size.

No Results Found is still 16px though. I think it should be that size, right?
cc. @wteouchicago

wteouchicago:
> > > Qn, what size should the "No properties with data" text be? It's a bit bigger than the mockup. Looks ok to me, but I'm not an expert on this. @cemile-ctds ?
> > 
> > 
> > The class we should use for the text size is `text-sm`.
> 
> I have updated the font size.
> 
> No Results Found is still 16px though. I think it should be that size, right? cc. @wteouchicago

Looks good! I think No results found at 16px makes sense -- it's the same size as the results.

amlehman (approved):

packages/portal-proto/src/features/cDave/Controls/ControlGroup.tsx
craigrbarnes: This is good.

craigrbarnes (approved):
Changes look fine to me and are an improvement.

PR #1329, opened 2024-10-03
PEAR-2087 - cDAVE: Investigate if analysis card's y-axis can display only integers 

amlehman:
## Description
Add option to only display integers on axis in bar chart

## Checklist

- [x] Added proper unit tests
- [x] Left proper TODO messages for any remaining tasks
- [x] Scanned for web accessibility with **aXe**, and mitigated or documented
      flagged issues

## Screenshots/Screen Recordings (if Appropriate)
Screenshot 2024-10-03 at 3 26 47 PM

PR #1604, opened 2025-08-15
PEAR-2453 - Change denominator for percentage calculation in Clinical Data Analysis

amlehman:
## Checklist

- [ ] Added proper unit tests
      flagged issues

### DIFF
--- a/packages/core/src/features/clinicalDataAnalysis/clinicalDataAnalysisSlice.ts
+++ b/packages/core/src/features/clinicalDataAnalysis/clinicalDataAnalysisSlice.ts
@@ -15,8 +15,15 @@ export const clinicalAnalysisApiSlice = endpointSlice.injectEndpoints({
         endpoint: "cases",
         fetchAll: false,
       }),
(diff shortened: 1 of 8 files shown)


wteouchicago:
Changes look good! 

I found a side effect that I am not sure can be fixed -- I can file a separate ticket as well:
- now that the denominator is all the cases in the cohort, we can run into situations where all the bars for the histogram in % of Cases mode are for very small percentages. When this happens, it looks like the y-axis has no tick labels (or whatever they are called).
- using a bladder cohort from home page, check out Cigarettes Per Day:

Screenshot 2025-08-19 at 11 13 04 AM

In https://github.com/NCI-GDC/gdc-frontend-framework/pull/1329, we made it so that the y-axis will only display integers. If we keep it as-is for the "# of Cases" view, but allow floats for the "% of Cases" view, would it allow the y-axis in this situation to have tick labels?

amlehman:
Hi, @wteouchicago. I took your suggestion and it seems to look good to me.

wteouchicago:
Works great! 👍

craigrbarnes (approved):

PR #1610, opened 2025-09-24
PEAR-2509: Fix for Repository - Percentages are not correct

paribartandhakal:
## Checklist

- [ ] Added proper unit tests
      flagged issues

### DIFF
--- a/packages/portal-proto/src/features/facets/types.ts
+++ b/packages/portal-proto/src/features/facets/types.ts
@@ -1,8 +1,7 @@
-import { FacetDefinitionType, GQLDocType, GQLIndexType } from "@gff/core";
+import { FacetDefinitionType, GQLDocType } from "@gff/core";
 
 export type FacetQueryOptions = {
-  readonly docType?: GQLDocType;
-  readonly indexType?: GQLIndexType;
+  readonly docType: GQLDocType;
   readonly facetType?: FacetDefinitionType;
 };
 
@@ -12,5 +11,4 @@ export interface CustomConfig {
   readonly handleCustomFilterSelected: (filter: string) => void;
   readonly handleResetCustomFilters: () => void;
   readonly defaultFilters: string[];
-  readonly queryOptions?: FacetQueryOptions;
 }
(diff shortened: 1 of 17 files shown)


packages/portal-proto/src/features/facets/types.ts
```diff
@@ -1,8 +1,7 @@
-import { FacetDefinitionType, GQLDocType, GQLIndexType } from "@gff/core";
+import { FacetDefinitionType, GQLDocType } from "@gff/core";
 
 export type FacetQueryOptions = {
-  readonly docType?: GQLDocType;
-  readonly indexType?: GQLIndexType;
```
amlehman, line 5: Why remove indexType here? I believe that is still being used in the hooks for custom facets.

amlehman:
Here's an example of using generics in portal-components for QueryOptions. I think it would be more robust typing than the current setup and enforce typing on the hooks as well. This definitely falls under optional/improvement and seems like it would be a decent size change so feel free to ignore. 

```
--- a/packages/portal-components/src/facets/FacetTabs.tsx
+++ b/packages/portal-components/src/facets/FacetTabs.tsx
@@ -21,7 +21,6 @@ import {
   CohortBuilderCategoryConfig,
   CustomFacetHooks,
   EnumChartProps,
-  QueryOptions,
 } from "./types";
 
 const StyledFacetTabs = (props: TabsProps) => {
@@ -65,19 +64,19 @@ const StyledFacetTabs = (props: TabsProps) => {
   );
 };
 
-type FacetGroupProps = {
+interface FacetGroupProps<Q> {
   readonly children?: React.ReactNode;
   readonly facets: FacetCardDefinition[];
   readonly hooks: FacetRequiredHooks;
-  readonly queryOptions?: QueryOptions;
+  readonly queryOptions?: Q;
 };
 
-export const FacetGroup: React.FC<FacetGroupProps> = ({
+export const FacetGroup = <Q,>({
   facets,
   children,
   hooks,
   queryOptions,
-}: FacetGroupProps) => {
+}: FacetGroupProps<Q>) => {
   const { usePopulateFacetData = () => {} } = hooks;
   usePopulateFacetData(facets, queryOptions);
 
@@ -94,17 +93,17 @@ export const FacetGroup: React.FC<FacetGroupProps> = ({
   );
 };
```

packages/portal-proto/src/features/facets/types.ts
paribartandhakal: I think I overlooked it part. Thank you

PR #1620, opened 2025-10-15
PEAR-2511: Fix for "Cohort has been saved" notification displayed even when /cohorts API fails

paribartandhakal:
## Description
Also, showing modals on delete cohort error, discard changes error.

## Checklist
- [ ] Added proper unit tests
      flagged issues


wteouchicago:
Works well for me! 👍

packages/portal-proto/src/features/cohortBuilder/CohortManager/cohortActionHooks.ts
```diff
@@ -452,6 +454,9 @@ export const useReplaceCohort = () => {
         .then((payload) => {
           updateCohortState({ payload, newName, cohortId, saveAs });
           result = { newCohortId: payload.id };
+        })
+        .catch((e) => {
```
amlehman, line 458: I get why the throw on line 407 is necessary, but are these other two needed?

packages/portal-components/src/cohort/CohortManager.tsx
```diff
@@ -162,11 +163,14 @@ const CohortManager: React.FC<CohortManagerProps> = ({
                         },
                       ]);
                   })
-                  .catch(
-                    () =>
-                      setCohortMessage &&
-                      setCohortMessage([{ cmd: "error", param1: "deleting" }]),
-                  );
+                  .catch(() => {
+                    modals.openContextModal({
```
amlehman, line 167: Nice 👍

packages/portal-proto/src/features/cohortBuilder/CohortManager/cohortActionHooks.ts
paribartandhakal: I just wanted to make sure that we propagate the backend error and show the error modal in every scenario.

packages/portal-components/src/modals/SaveCohortModal.tsx
```diff
@@ -114,9 +114,10 @@ const SaveCohortModal: React.FC<SaveCohortModalProps> = ({
         })
         .catch(() => {
           modals.openContextModal({
```
paribartandhakal, line 116: @amlehman this is where we catch it for replace.

packages/portal-proto/src/features/cohortBuilder/CohortManager/cohortActionHooks.ts
amlehman: You shouldn't need to catch and then throw the same error though. The default behavior is already that the exception will be thrown. You should only have a catch block if you are adding some additional error handling.

craigrbarnes (approved):

PR #1629, opened 2025-11-20
PEAR-2535 - Fix facet type for Cohort Builder cards

amlehman:
## Description
\+ PEAR-2530 - Fix max values for Cohort Builder cards

I also cleaned up the `facet_additional_data.json` file. It was a huge 4000 line fixture that we've only ever used for the range data 🥲 

## Checklist

- [x] Added proper unit tests
      flagged issues

Screenshot 2025-11-20 at 1 10 11 PM


wteouchicago:
1. I'm seeing a graphql 400 error with exposures.age_at_onset and also a min and max of 0.  It's fine on prod (but with the wrong units). `age_at_last_exposure` is OK. I also noticed `age_at_onset` is in the facet additional (range) data json while `age_at_last_exposure` isn't. 🤷‍♀️ 



```
{
    "data": {
        "viewer": {
            "explore": {
                "cases": {
                    "aggregations": null
                }
            }
        }
    },
    "errors": [
        {
            "locations": [
                {
                    "column": 9,
                    "line": 6
                }
            ],
            "message": "RequestError(400, 'search_phase_execution_exception', 'No [ranges] specified for the [exposures.age_at_onset:range] aggregation')"
        }
    ]
}
```

amlehman:
Hey @wteouchicago, this is basically an edge case of our existing bucket range code. The logic for the year buckets is `const numBuckets = Math.round((adjMaximum - adjMinimum) / 10)`. Since we have a max and min of zero that results in creating zero buckets. We then send no buckets to the range query which isn't a valid query resulting in the error. I think the best approach is probably to change the bucket creation code to always create at least one bucket?

craigrbarnes (approved):
Much better than the original 💯

wteouchicago:
> Hey @wteouchicago, this is basically an edge case of our existing bucket range code. The logic for the year buckets is `const numBuckets = Math.round((adjMaximum - adjMinimum) / 10)`. Since we have a max and min of zero that results in creating zero buckets. We then send no buckets to the range query which isn't a valid query resulting in the error. I think the best approach is probably to change the bucket creation code to always create at least one bucket?

@amlehman I think I'm missing something -- why do we have a max and min of 0 for this property?

amlehman:
Oh wait, great question @wteouchicago 😆 . The year facet code is assuming that the max/min we pass in needs to be converted from days into years: https://github.com/NCI-GDC/gdc-frontend-framework/blob/develop/packages/portal-components/src/facets/facetTypes/NumericRangeFacet/Years.tsx#L19

So 0 - 89 gets converted to 0 - 0

PR #1636, opened 2025-12-04
Qa 2416 upgrade holmes py python version

bpierce01:
## Description
- Upgraded holmes-py to use python 3.12.x


bpierce01:
The protobuf vulnerability is not new and was not introduced in this PR. I have made a ticket on the QA board to address it: https://gdc-ctds.atlassian.net/browse/QA-2547

holmes-py/.python-version
```diff
@@ -1 +1 @@
-3.9.4
+3.12.1
```
uc-pperumala: Do we want this to be `3.12` allowing users to use any flavor of 3.12.x or more strictly `3.12.1`?

uc-pperumala: I would do `3.12`, if it works (it appears to be working for me).

uc-pperumala (approved):
Left a minor comment to update `.python-version` to include just `3.12` so that the users can use any flavor of 3.12. Otherwise, the setup works just fine. Approved.

holmes-py/.python-version
bpierce01: good call, updated file

craigrbarnes (changes requested):
The security pipeline is failing on the requirement.txt file.

PR #1543, opened 2025-05-14
PEAR-2287: Deleting cohorts while in Cohort Comparison

paribartandhakal:
## Checklist

- [ ] Added proper unit tests
- [ ] Left proper TODO messages for any remaining tasks
- [ ] Scanned for web accessibility with **aXe**, and mitigated or documented
      flagged issues

PR #1651, opened 2026-01-05
PEAR-2510 - Cohort Comparison - Replacing active cohort on analysis screen takes user back to selection screen

amlehman:
## Description
This issue was caused by the [fix](https://github.com/NCI-GDC/gdc-frontend-framework/pull/1543) to kick a user back to the selection screen if they deleted a cohort. We've had a lot of issues with these cohort selection screens and changing/updating/deleting the cohorts so I wanted to approach the fix in a different way. Since cohort IDs can change for a multitude of reasons, we can't really accurately infer what has happened to the cohort from just whether it is in the list of cohort IDs. So I tried the approach of adding an explicit property to the cohort.  

## Checklist

- [ ] Added proper unit tests
      flagged issues


packages/portal-proto/src/features/cohortBuilder/CohortManager/cohortActionHooks.ts
```diff
@@ -74,7 +75,7 @@ export const useDeleteCohort = () => {
 
   const [deleteCohortFromBE] = useDeleteCohortMutation();
   const deleteCohort = useDeepCompareCallback(() => {
-    coreDispatch(removeCohort({}));
+    coreDispatch(queueCohortForRemoval({}));
```
paribartandhakal: Why do we need to queue cohort for removal here and not remove it instantly like before?

wteouchicago:
I tested deletion-and-replacement-while-in-an-app scenarios pretty thoroughly. Everything looks good, but there is one change in behaviour:
- in Set Operations, if you delete cohorts being compared, you are no longer kicked to the Selection Screen. 

Is this an easy fix? If not, we can think about whether this new behaviour is ok/justifiable.

packages/portal-proto/src/features/cohortBuilder/CohortManager/cohortActionHooks.ts
amlehman: So that we have a way to tell when the cohort has been removed. See this comment :)

`This issue was caused by the fix to kick a user back to the selection screen if they deleted a cohort. We've had a lot of issues with these cohort selection screens and changing/updating/deleting the cohorts so I wanted to approach the fix in a different way. Since cohort IDs can change for a multitude of reasons, we can't really accurately infer what has happened to the cohort from just whether it is in the list of available cohort IDs. So I tried the approach of adding an explicit property to the cohort.`

paribartandhakal: Got it. I was confused earlier and thought that removal can happen from multiple places. Now I remember that cohort removal only happens from the cohort bar. This makes sense.

amlehman:
Hi @wteouchicago, I think on set operations you only get kicked back to the selection screen if you delete a cohort and you no longer have enough cohorts selected to compare. Has that behavior changed? It's looking the same to me between local and prod.

wteouchicago:
> Hi @wteouchicago, I think on set operations you only get kicked back to the selection screen if you delete a cohort and you no longer have enough cohorts selected to compare. Has that behavior changed? It's looking the same to me between local and prod.

Oops, you're right, @amlehman -- I had only two cohorts on prod and qa-int when I tested earlier, so deleting one meant I was always kicked to the selection screen.

On local, if I start with 3 cohorts:
- deleting 1 cohort didn't update the analysis page
- deleting 2+ cohorts didn't kick me to the selection page

wteouchicago:
works great! 👍

packages/portal-proto/src/features/cohortBuilder/CohortManager/cohortActionHooks.ts
paribartandhakal: @amlehman I noticed that `removeCohort` is still used in a few places. Is this intended? 

Should we use `queueCohortForRemoval` everywhere now, or is there a distinction between the two?

I see `removeCohort` used here:

1. `useSetupInitialCohorts`
2. `useUpdateCohortState`
3. `useReplaceCohort`

amlehman: Yeah, that was intentional. I meant that action only for the act when the **user** deletes a cohort (not when we remove the cohort from the store for some reason). I was struggling with what to name these actions to convey that. Any suggestions?

wteouchicago: `userDeletesCohort`? :P

paribartandhakal: Hah ya this sounds good! :D

paribartandhakal (approved):
All looks good. Just would leave a comment to distinguish between the two removal functions for down the road.

packages/portal-proto/src/features/cohortBuilder/CohortManager/cohortActionHooks.ts
amlehman: Okay, I've updated the names of these actions and their comments. Hopefully it is more clear when to use what action.

craigrbarnes (approved):

PR #1652, opened 2026-01-05
PEAR-2498 - Protein change and consequences discrepancies

amlehman:
## Checklist

- [ ] Added proper unit tests
      flagged issues

### DIFF
--- a/packages/portal-proto/src/features/manageSets/SetDetailPanel.tsx
+++ b/packages/portal-proto/src/features/manageSets/SetDetailPanel.tsx
@@ -94,18 +94,25 @@ const SetDetailPanel: React.FC<SetDetailPanelProps> = ({
           : [];
       } else {
         return isMutationSuccess && !isMutationFetching
-          ? ssmsDetailData?.hits?.map((ssm) => ({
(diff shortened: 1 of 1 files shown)


wteouchicago:
1. crashed when my mutation set consisted of only this mutation with id `02822cde-20cb-5770-9c7c-a254c6f3484c` and I went to Manage Sets to view the set's contents.
Screenshot 2026-01-05 at 5 35 02 PM
- note that this mutation has no canonical transcript.

Other areas looked good.

amlehman:
@wteouchicago, what should be displayed in the consequence column in that case? An empty string?

wteouchicago:
> @wteouchicago, what should be displayed in the consequence column in that case? An empty string?

Please display "--" in that situation (I'll update my comment in the ticket). It turns out we just display "--" for Protein Change in Mutation Frequency for this situation, so we can just be consistent with that.

wteouchicago:
looks great! 👍 ❗ 👍

paribartandhakal (approved):

craigrbarnes (approved):

PR #1662, opened 2026-01-27
PEAR-2594 - Summary page not loading until refresh if user agreement displayed and dismissed

amlehman:
## Description
Adding the loading state for fetching the data version is the fix here, the rest is just cleaning up some prototype code that was still laying around 🙀 

This bug is tied to this [serialization code](https://github.com/NCI-GDC/gdc-frontend-framework/blob/develop/packages/core/src/serializeQueryArgs.ts) that adds the data version to the rtk query cache key. The modal ended up having nothing to do with the bug, it occurs when there's no data already cached and you go to a page that uses one of these [hooks](https://github.com/NCI-GDC/gdc-frontend-framework/blob/develop/packages/core/src/features/gdcapi/index.ts#L219). The request was getting 'stuck' in the fetching state because the query cache key doesn't match from when the request is fired vs completed. 

## Checklist

- [ ] Added proper unit tests
      flagged issues


wteouchicago:
Oops, I missed looking at this.

Works great for me -- all the entity pages that were not loading previously are now loading as expected.

paribartandhakal (approved):

craigrbarnes (approved):

PR #1669, opened 2026-02-07
update proteinpaint-client to 2.170.8

siosonel:
## Description

Correlation Plot fixes:
- [SV-2725](https://gdc-ctds.atlassian.net/browse/SV-2725)
- [SV-2726](https://gdc-ctds.atlassian.net/browse/SV-2726)
- [SV-2727](https://gdc-ctds.atlassian.net/browse/SV-2727)
- [SV-2728](https://gdc-ctds.atlassian.net/browse/SV-2728)
- [SV-2731](https://gdc-ctds.atlassian.net/browse/SV-2731)
- [SV-2734](https://gdc-ctds.atlassian.net/browse/SV-2734)
- [SV-2735](https://gdc-ctds.atlassian.net/browse/SV-2735)
- [SV-2736](https://gdc-ctds.atlassian.net/browse/SV-2736)
- [SV-2737](https://gdc-ctds.atlassian.net/browse/SV-2737)

From PP release changelog:
- remove all untagged html text in menu.clear(), previously only tagged children were removed
- hide not-fully functional undo-redo buttons in plot and summary components
- make the case count limit message applicable to both correlation plot and gene exp clustering
- remove 5K case limit when doing hiercluster; avoid 1min /cases/ query on empty cohort filter
- lessen GDC API requests by generating sample filter only once in mayGetGeneVariantData()
- add max limit to number of genes supplied to geneVariant gene set
- prevent the submission of plot edits while input variable data is still being loaded
- downgrade missing plot config into a warning, instead of displaying an irrelevant error message
- remove add-to-filter option from correlation plot selection of rendered data; fix the returned GDC API  bins from getData()
- support correlation plot triggerAbort() to cancel stale network requests

## Checklist

- [ ] Added proper unit tests
      flagged issues

### DIFF
--- a/packages/portal-proto/src/features/proteinpaint/CorrelationWrapper.tsx
+++ b/packages/portal-proto/src/features/proteinpaint/CorrelationWrapper.tsx
@@ -40,8 +40,8 @@ export const CorrelationWrapper: FC<PpProps> = (props: PpProps) => {
     : buildCohortGqlOperator(currentCohort);
   const userDetails = useFetchUserDetailsQuery();
   const prevData = useRef<any>();
+  const toolApp = useRef<any>();
   const [isLoading, setIsLoading] = useState(false);
-
   const showLoadingOverlay = () => setIsLoading(true);
   const hideLoadingOverlay = () => setIsLoading(false);
   const matrixCallbacks: RxComponentCallbacks = {
@@ -111,10 +111,25 @@ export const CorrelationWrapper: FC<PpProps> = (props: PpProps) => {
           // in case of race condition
           return prevData.current != data;
         },
-      });
+      })
+        .then?.((_app) => {
+          toolApp.current = _app;
+        })
+        .catch((e) => {
+          // the app should either work or display an error in a red banner within the tool container div,
+          // this uncaught-by-app error is unlikely to happen except for bundling issues that are not detected at build time
+          console.error(e);
+        });
+
+      return () => {
+        if (!toolApp.current || window.location.href.includes("Correlation"))
+          return;
+        // cancel unnecessary network requests when this tool app is hidden
+        toolApp.current.triggerAbort();
+      };
     },
     // eslint-disable-next-line react-hooks/exhaustive-deps
-    [filter0, userDetails],
+    [filter0, userDetails?.data],
   );
 
   const divRef = useRef();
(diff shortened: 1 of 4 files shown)


packages/portal-proto/src/features/proteinpaint/CorrelationWrapper.tsx
```diff
@@ -111,10 +111,25 @@ export const CorrelationWrapper: FC<PpProps> = (props: PpProps) => {
...
         },
-      });
+      })
+        .then?.((_app) => {
+          toolApp.current = _app;
+        })
+        .catch((e) => {
+          // the app should either work or display an error in a red banner within the tool container div,
+          // this uncaught-by-app error is unlikely to happen except for bundling issues that are not detected at build time
+          console.error(e);
```
amlehman: I'm not sure I understand what this comment is getting at but we should be avoiding directly logging errors to console in the GDC.

siosonel: I can remove the `console.log()` or change to a `console.warn()` or display the error inside an error div in the `CorrelationWrapper` and `MatrixWrapper`. 

To explain the comment, we do most of our testing in dev and locally packed `proteinpaint-client` package as used in non-GDC portals.. For GDC, we do another round of tests after `proteinpaint-client` is rebundled as part of GFF. This rebundling may introduce some rare untested scenarios, where some webpack/bundler config somehow affects proteinpaint-based features or functionality. Almost all errors should be caught inside PP components or bubble up to the app API, and be shown in a red banner or intentionally ignored. The `console.error()` in line 121 is only for the rare case that there is an error that somehow didn't get caught within the app - it is assumed it's not related to tool functionality because it if did, then it would have been caught inside the app. These non-app/tool error would likely not be meaningful to an end-user, so by logging in the console, a developer can still be informed that an error occured.

siosonel: Also, I've seen other errors/warnings/info in the console when viewing the portal, so I thought it wasn't unusual to use console.error() in the rare case that I mentioned above.
 
Screenshot 2026-02-09 at 11 48 31 AM

philtom-ctds: I agree with Amy. We should avoid logging to console in production. If you **really** need this, then I'd recommend a sparing use of `console.debug` instead.

Those network errors are coming from `fetch` itself.

siosonel: ok, I removed the `.catch()` callback

amlehman (approved):

philtom-ctds (approved):

PR #1674, opened 2026-02-20
PEAR-2555 - Update React for Portal 2.0

amlehman:
## Description
Big React changes:
- JSX was removed from the global namespace in react/types 
- useRef requires an argument

Libraries with React 19> peer dependencies removed:
- react-tabs (not used, probably replaced by mantine?)
- react-tour (tour wasn't ever released)
- react-table (we were already using the newer library @tanstack/table)

Other stuff:
- I had to update redux/redux-toolkit with this. They added a new optional parameter to the thunk [method signature](https://github.com/reduxjs/redux-toolkit/blob/master/packages/toolkit/src/createAsyncThunk.ts#L268) that resulted in the type changes in core. 
- I moved React to the top level of package.json only. I can move this back if we prefer each package to have React as a dependency. It was just easier to update it in one place and assure that there were no conflicting versions. 

## Checklist

- [ ] Added proper unit tests
      flagged issues


wteouchicago:
@amlehman any specific areas we should focus on for testing? Or is it basically...everything?

packages/core/src/coreCreateApi.ts
```diff
@@ -14,7 +14,7 @@ const coreCreateApi: CreateApi<keyof ApiModules<any, any, any, any>> =
       hooks: {
         useSelector: useCoreSelector as any,
         useStore: useCoreStore as any,
-        useDispatch: useCoreDispatch,
+        useDispatch: useCoreDispatch as any,
```
amlehman, line 17: This is addressed in https://github.com/NCI-GDC/gdc-frontend-framework/pull/1675

amlehman:
@wteouchicago, basically everything 🫣

amlehman:
I guess to clarify, the files with just import changes shouldn't need any testing. The files with changes to `useRef` could potentially have changes to behavior. They aren't really contained to a specific part of the site though. I could make a list of those components but I think the best way to test is still probably going to be 'click through the site'.

wteouchicago:
> I guess to clarify, the files with just import changes shouldn't need any testing. The files with changes to `useRef` could potentially have changes to behavior. They aren't really contained to a specific part of the site though. I could make a list of those components but I think the best way to test is still probably going to be 'click through the site'.

@amlehman no problem, I can 👀 and 🙏 and click around.

craigrbarnes (approved):
A large number of changes and everything appears fine to me.

package.json
```diff
@@ -61,11 +60,14 @@
     "@mocks-server/main": "^4.1.0",
     "@next/eslint-plugin-next": "^12.1.5",
     "@svgr/webpack": "^8.0.1",
+    "@swc/core": "1.9.2",
```
paribartandhakal, line 63: I am not  sure if we are even using this. I don't see it being used.

paribartandhakal: looks like Next JS installs on this own, so we might not need it anymore.

paribartandhakal (approved):
Looks good to me. Most of the hook structure in core looks the same, but i think it's okay to redeclare without coupling it too much.

I hope they had codemod to add JSX.

PR #1682, opened 2026-03-02
IDC viewer app

aacic:
## Description

This PR adds the IDC images viewer app to GDC. 

The app works in a way that, on load, it downloads the parquet file, which contains IDC - GDC mappings. The mapping logic is explained in the [GDC-GDC-IDCImageViewerUserStories-171125-1555-6.pdf](https://github.com/user-attachments/files/25848650/GDC-GDC-IDCImageViewerUserStories-171125-1555-6.pdf)

To test:

1. Set a cohort
2. Open the IDC app 
3. Verify that the mapping works as explained in [GDC-GDC-IDCImageViewerUserStories-171125-1555-6.pdf](https://github.com/user-attachments/files/25848650/GDC-GDC-IDCImageViewerUserStories-171125-1555-6.pdf)

## Checklist

- [x] Added proper unit tests
      flagged issues


packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
```diff
@@ -0,0 +1,588 @@
...
+  }, [cohortGqlFilters, loadIdcData, mapSortingToSortBy]);
+
+  useEffect(() => {
+    fetchAllGdcCasesAndMap();
+    // intentionally only depend on sorting and the stable cohortFiltersKey
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [sorting, cohortFiltersKey]);
+
+  useEffect(() => {
+    const root = idcViewerRootDivRef.current;
```
amlehman: It doesn't seem necessary to directly manipulate the DOM here, why not just add this style to the div that this ref is pointing at?

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
```diff
@@ -0,0 +1,588 @@
...
+      };
+
+      const extendedFilters = caseFiltersArg
+        ? mergeWithAndFilters(caseFiltersArg, id_filters)
+        : id_filters;
+
+      const sortByForApi = mapSortingToSortBy();
+
+      // Fetch all matching GDC cases in a single request (request size = number of case_ids)
+      const resp = await fetchGdcCasesApi({
```
amlehman: I would recommend using the `useGetCasesQuery` hook instead of calling this directly. You are recreating a lot of the logic (refetching when the input changes, loading and error states, blocking requests while one is in flight) that the hook will handle for you. Here's an [example](https://github.com/NCI-GDC/gdc-frontend-framework/blob/develop/packages/portal-proto/src/features/cases/CaseSummary.tsx#L21) of that hook being used.

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
```diff
@@ -0,0 +1,588 @@
...
+    expandedCases.forEach((id) => {
+      o[id] = true;
+    });
+    return o;
+  }, [expandedCases]);
+  // NEW: global loading state for parquet download/parsing + GDC fetch
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+  // helper to toggle when VerticalTable calls setExpanded(row, columnId)
+  const setTableExpanded = useCallback(
+    (row: any /* tanstack Row */) => {
```
amlehman: tanstack has a Row type defined: `import { Row } from "@tanstack/react-table";`

packages/portal-proto/src/features/idc/IDCViewerWrapper.unit.test.tsx
```diff
@@ -0,0 +1,119 @@
...
+    </MantineProvider>,
+  );
+
+  const gdcCaseIDEl = await findByText("GDC Case ID");
+  expect(gdcCaseIDEl).toBeInTheDocument();
+
+  const programEl = await findByText("Program");
+  expect(programEl).toBeInTheDocument();
+
+  unmount();
```
amlehman: It isn't necessary to manually call unmount when using react testing library with jest: https://testing-library.com/docs/react-testing-library/api/#cleanup

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
```diff
@@ -0,0 +1,588 @@
...
+  const [sorting, setSorting] = useState<SortingState>([
+    { id: "caseId", desc: false },
+  ]);
+
+  // use current cohort filters (hook)
+  const currentCohortFilterSet = useCurrentCohortFilters();
+
+  // Convert FilterSet -> Operation, but derive it from a stable JSON key so the
+  // effect below doesn't retrigger due to changing object identities from the hook.
+  const cohortFiltersKey = useMemo(
```
amlehman: We have been using the `use-deep-compare` library which does a deep comparison on the contents of the dependency array, I think using that would eliminate the need for this 'key'

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
```diff
@@ -0,0 +1,588 @@
...
+  "StudyDate",
+  "StudyDescription",
+  "study_type",
+  "gdc_case_id",
+];
+
+// Helper: read idc_data from a parquet file
+async function readParquetIndex(
+  idc_index_file: any,
+): Promise<{ idc_data: ReadonlyArray<any>; case_ids: readonly string[] }> {
```
amlehman: Is the data in the file at least partially fixed/known? I think defining a Typescript interface for this data would help the readability of the code.

wteouchicago:
@aacic / @siosonel -- should we wait for SJ to finish reviewing before the UC team does their review?

siosonel:
> @aacic / @siosonel -- should we wait for SJ to finish reviewing before the UC team does their review?

@wteouchicago You don't have to wait for SJ review. I'll add you and Craig to the reviewers.

wteouchicago:
> > @aacic / @siosonel -- should we wait for SJ to finish reviewing before the UC team does their review?
> 
> @wteouchicago You don't have to wait for SJ review. I'll add you and Craig to the reviewers.

Hi @siosonel -- I spoke with the team and we think it's better if UC only reviews after SJ has reviewed. That way, UC is not reviewing something that has not been finalized by your team.

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
```diff
@@ -0,0 +1,588 @@
...
+                                  </td>
+                                  <td style={{ padding: 6 }}>
+                                    {radioLink ? (
+                                      <a
+                                        href={radioLink}
+                                        target="_blank"
+                                        rel="noreferrer"
+                                        onClick={(e) => e.stopPropagation()}
+                                      >
+                                        Open study
```
siosonel: It's not obvious that `Open study` is clickable, and I'm not sure if that's necessary or how to do so using existing portal css styles/classes- may need guidance from @craigrbarnes  or @wteouchicago .

Screenshot 2026-03-13 at 12 13 20 PM

packages/portal-proto/src/features/user-flow/workflow/registeredApps.tsx
```diff
@@ -245,6 +246,26 @@ export const REGISTERED_APPS: AppRegistrationEntry[] = [
...
+        style={{ margin: "0" }}
+        aria-hidden="true"
+      />
+    ),
+    tags: ["IDCViewer"],
+    hasDemo: false,
+    description:
+      "View histology and radiology images using the Imaging Data Commons (IDC) Image Viewer for cohort cases.",
+    id: "IDCViewerApp",
+    countsField: "caseCount",
```
siosonel, line 265: I don't see case counts in the card. I think this may require a custom case or file count hook similar to `useLazyScRNAseqCaseCountQuery` - @craigrbarnes please advise.

Image

siosonel:
There is a race condition with TCGA data loading after I have already changed to the ALCHEMIST saved cohort. An abort signal may need to be passed somewhere to cancel stale cohort data request, or maybe pass cohortFiltersKey to one of the `useEffects` that returns a clean up function - may need guidance from @craigrbarnes . 

Screenshot 2026-03-13 at 11 58 19 AM

siosonel:
@wteouchicago I've now reviewed the changes based on acceptance criteria that I'm familiar with, such as required from my own GFF experience and also from the linked pdf. There are image ID mapping requirements and GFF coding patterns that I leave for others to review.

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
wteouchicago: Dev would know more of the code details, but if it's opening an external site, then elsewhere in the portal, we include an external site icon (see the External References in the gene summary page for examples of what I mean). The text is underlined to show that it's a clickable link.

Screenshot 2026-03-13 at 12 20 17 PM

packages/portal-proto/src/features/user-flow/workflow/registeredApps.tsx
wteouchicago: I'm not as involved in the IDC image viewer discussions so I may have missed this. Is it basically all cases that have slide images in the GDC that can be used with the app?

siosonel: It's from the IDC parquet file, per one of the acceptance criteria in the pdf.

Screenshot 2026-03-13 at 12 27 28 PM

wteouchicago:
@aacic / @siosonel - is there something special I need to do to run the new tool? I'm seeing:

Screenshot 2026-03-13 at 4 22 11 PM

siosonel:
> @aacic / @siosonel - is there something special I need to do to run the new tool? I'm seeing:
> 
> Screenshot 2026-03-13 at 4 22 11 PM

I didn't do anything special, just my usual `npm run dev` that calls a ProteinPaint server instance for SJ tools, but that should not be relevant to the IDC image viewer. I loaded https://localhost.gdc.cancer.gov:3010/analysis_page?app=IDCViewerApp just now and still get a fully rendered table, using `TCGA` cohort. Maybe your cohort is too big or too slow to load?

wteouchicago:
> I didn't do anything special, just my usual `npm run dev` that calls a ProteinPaint server instance for SJ tools, but that should not be relevant to the IDC image viewer. I loaded https://localhost.gdc.cancer.gov:3010/analysis_page?app=IDCViewerApp just now and still get a fully rendered table, using `TCGA` cohort. Maybe your cohort is too big or too slow to load?

Oh I see, @siosonel . I'll use the https://localhost.gdc.cancer.gov:3010/ method (I was using the localhost:3000 method earlier).

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
craigrbarnes: For this links, you can use the `AnchorLink` component for consistency:
 `packages/portal-proto/src/components/AnchorLink.tsx`

packages/portal-proto/src/features/user-flow/workflow/registeredApps.tsx
craigrbarnes: This will need a custom count query hook, or possibly an extension of an existing query, to get the correct counts.

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
craigrbarnes: @amlehman is correct, I would `useGetCasesQuery`

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
craigrbarnes: Agreed. We should try to avoid the use of `any` if possible.

packages/portal-proto/src/components/Table/ExpandRowComponent.tsx
```diff
@@ -1,49 +1,53 @@
-import React, { JSX } from "react";
+import React from "react";
 import { CollapseCircleIcon, ExpandCircleIcon } from "@/utils/icons";
 
 const ExpandRowComponent = ({
   isRowExpanded,
   value,
   isColumnExpanded = true,
   title,
+  handleOneElementValue = true,
```
craigrbarnes: This name is slightly vague. It sounds like a function, not a boolean flag.
Something clearer:

- showSingleValueDirectly
- collapseSingleValue
- renderSingleValueInline

Not a bug, just readability debt.

packages/portal-proto/src/components/Table/ExpandRowComponent.tsx
```diff
@@ -1,49 +1,53 @@
...
-    )}
-  </>
-);
+  handleOneElementValue?: boolean;
+}): React.ReactNode => {
+  if (value.length === 0) {
+    return "--";
+  }
+
+  if (value.length === 1 && handleOneElementValue) {
```
craigrbarnes: This will create an inconsistency in the UI. If I read this code correctly, when the value. length === 1 && handleOneElementValue is false, it will render 1 <icon> "value". This is not consistent with how single values are rendered in other tables across the site.

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
```diff
@@ -0,0 +1,588 @@
...
+          const wsiCount = studies.filter((s) => s?.hasWSI).length;
+          const radiologyCount = studies.filter((s) => s?.hasRadiology).length;
+
+          const idcNoun = count === 1 ? "IDC study" : "IDC studies";
+          const wsiLabel = `${wsiCount} WSI`;
+          const radioLabel = `${radiologyCount} Radiology`;
+          const titleLabel = `${idcNoun} (${wsiLabel} + ${radioLabel})`;
+
+          if (count === 0) {
+            return <span style={{ color: "#666" }}>-</span>;
```
craigrbarnes: You need to use the colors defined in Tailwind and use tailwind to style tis

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
```diff
@@ -0,0 +1,588 @@
...
+    >
+      <div style={{ marginTop: 12 }}>
+        <div style={{ maxHeight: 460, overflow: "auto" }}>
+          <div style={{ position: "relative", minHeight: 120 }}>
+            <LoadingOverlay visible={isLoading} zIndex={50} />
+            {!isLoading &&
+              (loadError ? (
+                <div
+                  data-testid="idc-error-message"
+                  style={{ padding: 16, color: "#900", fontSize: 14 }}
```
craigrbarnes: Again use tailwind

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
```diff
@@ -0,0 +1,588 @@
...
+                >
+                  There was an error rendering IDC table, please try again
+                  later.
+                </div>
+              ) : tableData.length === 0 ? (
+                <div
+                  data-testid="no-idc-message"
+                  style={{ padding: 16, color: "#666", fontSize: 14 }}
```
craigrbarnes: Tailwind

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
```diff
@@ -0,0 +1,588 @@
...
+  useEffect(() => {
+    fetchAllGdcCasesAndMap();
+    // intentionally only depend on sorting and the stable cohortFiltersKey
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [sorting, cohortFiltersKey]);
+
+  useEffect(() => {
+    const root = idcViewerRootDivRef.current;
+    if (!root) return;
+    root.style.background = "#fff";
```
craigrbarnes: There are utility colors in Tailwind that should be used.

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
```diff
@@ -0,0 +1,588 @@
...
+            </div>
+          );
+        },
+        enableSorting: false,
+      },
+    ],
+    [setTableExpanded],
+  );
+
+  // Use client-side pagination hook (reads page size from VerticalTable via handleChange)
```
craigrbarnes: If pagination is handled client-side, this could get quite large, considering the number of cases in the GDC. If you use @amlehman's recommendation to use `useGetCasesQuery`, pagination can be handled on the server.

packages/portal-proto/next.config.js
```diff
@@ -11,6 +11,7 @@ const connectSrc = [
   "https://www.google-analytics.com",
   "https://dap.digitalgov.gov",
   "https://metrics.cancer.gov",
+  "https://storage.googleapis.com",
```
philtom-ctds: This allows access to all data in Google Cloud Storage, which is too broad.  Narrow this policy directive.  Ideally, it would be the specific object that will be read. If the url is not static, then you'll need to work with IDC to determine the narrowest policy.

package.json
```diff
@@ -46,6 +46,7 @@
     "dom-to-svg": "^0.12.2",
     "echarts": "^5.5.1",
     "file-saver": "^2.0.5",
+    "hyparquet": "^1.23.0",
```
philtom-ctds: This looks off to me.  If it's only used in the portal-proto workspace, then declaring the dependency in the portal-proto package.json should be sufficient and not needed here.  See the README.md for instructions on how to add a dependency to a specific workspace.

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
aacic: Done.

packages/portal-proto/src/features/user-flow/workflow/registeredApps.tsx
aacic: @craigrbarnes @siosonel @wteouchicago The spec is a bit outdated. Himanso and Xin decided not to display case counts. I'll leave it like it is for now, and I'll double-check this with Himanso again.

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
aacic: Removed.

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
aacic: Done.

packages/portal-proto/src/features/idc/IDCViewerWrapper.unit.test.tsx
aacic: Removed.

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
aacic: Done.

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
aacic: Done.

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
aacic: Done.

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
aacic: Done.

packages/portal-proto/src/components/Table/ExpandRowComponent.tsx
aacic: Renamed to renderSingleValueInline.

packages/portal-proto/src/components/Table/ExpandRowComponent.tsx
aacic: Let me just explain what I hoped to achieve here.

Before my change, one value row was rendered like this:

Project

So, it was not possible to expand it. Please look at the last row. My change this break this behavior. 

By adding the option: renderSingleValueInline = false (it's true by default)
The one-value rows are rendered like this:
IDC viewer

So, it's possible to render one value row as required by the spec.

packages/portal-proto/src/features/user-flow/workflow/registeredApps.tsx
```diff
@@ -245,6 +246,26 @@ export const REGISTERED_APPS: AppRegistrationEntry[] = [
     noDataTooltip:
       "Current cohort does not have scRNAseq data available for visualization.",
   },
+  {
+    name: "IDC Image Viewer",
+
```
paribartandhakal: nitpick: extra newline space

packages/portal-proto/next.config.js
aacic: The URL is static, I've set it to: 

`https://storage.googleapis.com/idc-index-data-artifacts/current/release_artifacts/gdc_idc_mapping.parquet`

package.json
paribartandhakal: Yeah, I can see that already being installed in `portal-proto`

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
```diff
@@ -0,0 +1,582 @@
+// File: `packages/portal-proto/src/features/proteinpaint/IDCViewerWrapper.tsx`
```
paribartandhakal: Nitpick: We don't usually mention this.

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
```diff
@@ -0,0 +1,582 @@
...
+                  data-testid="no-idc-message"
+                  className="p-4 text-gdc-grey-dark text-sm"
+                >
+                  No idc images for selected cohort.
+                </div>
+              ) : (
+                // VerticalTable-based view
+                (() => {
+                  // renderSubComponent: show detailed studies list for expanded row
+                  const renderSubComponent = ({ row }: any) => {
```
paribartandhakal: It is much better if it can be defined outside the return, or in its own component.

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
```diff
@@ -0,0 +1,582 @@
...
+  const setTableExpanded = useCallback(
+    (row: Row<string>) => {
+      // row.id is computed by getRowId (we will use caseId)
+      if (!row?.id) return;
+      toggleExpanded(row.id);
+    },
+    [toggleExpanded],
+  );
+
+  const buildSlimStudyURL = (studyInstanceUID: string) =>
```
paribartandhakal: It can be defined outside the component (module level) as it doesn't change.

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
```diff
@@ -0,0 +1,582 @@
...
+
+  // Build pagination object for VerticalTable / TablePagination using hook values
+  const pagination = {
+    size,
+    page,
+    pages,
+    from: from,
+    total,
+    label: "study",
+    customPluralLabel: "study",
```
paribartandhakal: This should be studies.

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
```diff
@@ -0,0 +1,582 @@
...
+        programName,
+        studiesList,
+        studiesCount: studiesList.length,
+        wsiCount,
+        radiologyCount,
+      };
+    });
+  }, [mappings]);
+
+  const columns: ColumnDef<any>[] = useMemo(
```
paribartandhakal: Consider using `createColumnHelper` with a typed row interface instead of` ColumnDef<any>` — see [AnnotationsTable](https://github.com/NCI-GDC/gdc-frontend-framework/blob/develop/packages/portal-proto/src/features/cases/AnnotationsTable.tsx#L47) for the established pattern in this codebase.

packages/portal-proto/src/components/Table/ExpandRowComponent.tsx
```diff
@@ -1,49 +1,53 @@
...
-  </>
-);
+  renderSingleValueInline?: boolean;
+}): React.ReactNode => {
+  if (value.length === 0) {
+    return "--";
+  }
+
+  if (value.length === 1 && renderSingleValueInline) {
+    return value[0];
```
wteouchicago: qn: would `value` instead of `value[0]` here work too?

packages/portal-proto/src/components/Table/ExpandRowComponent.tsx
```diff
@@ -1,49 +1,53 @@
-import React, { JSX } from "react";
+import React from "react";
 import { CollapseCircleIcon, ExpandCircleIcon } from "@/utils/icons";
 
 const ExpandRowComponent = ({
   isRowExpanded,
   value,
   isColumnExpanded = true,
   title,
+  renderSingleValueInline = true,
```
wteouchicago: I'm trying to figure out if someone using our table component in the future would understand what `renderSingleValueInline` does without referring to the discussion here in this PR. Separately, we may want to include some information about this in our docs, so I want to make sure it's explainable.

I'm thinking out loud here:
- our table component is designed to follow a fairly common standard UX practice (not just in the portal) where we hide away (in the expanded row) things that most users would not need to see all the time (e.g. SSMs broken down by projects in Mutation Frequency or the full list of 20 disease types a project contains). This reduces clutter and ideally doesn't cause users to need to click all the time to see things they need to see/use. I'll think of this as the default behavior since it was already the standard in the portal.
- your use case here doesn't quite follow that pattern. If I understand correctly, the expanded row is intended to be used whether a case has just one IDC study or many, and the expanded row contains another table. (As an aside, in almost all situations, I only spotted 1 study per case and I didn't see any examples in the specs on how cases with 1 study were supposed to be handled). It's also possibly hiding away some important info and link(s) for the user in the expanded row. So it's kind of an edge case.

So:
1. this is a qn for @craigrbarnes / @amlehman / @paribartandhakal -- is there any common practice in the GFF for situations like this for whether it's the default situation or the edge case that drives the naming of the flag, i.e. do we normally have `doDefault = true` or `doEdgeCase = true`? If so, for clarity, we should follow the pattern here.
2. this may end up being affected by (1). I didn't find `renderSingleValueInline` clear enough based on the explanation of the intention here (https://github.com/NCI-GDC/gdc-frontend-framework/pull/1682#discussion_r2960585280). For example, would someone know that "inline" is the main (and not the expanded) row? Depending on how we go with (1), maybe we can try something like `displayAllValuesInExpandedRow`, `displayNestedTable`, etc.?
3. can a comment be added in the code that explains what the flag does? (We can then use it in the docs if needed.)

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
aacic: Fixed.

packages/portal-proto/src/features/user-flow/workflow/registeredApps.tsx
aacic: Fixed.

package.json
aacic: Removed from here, the new deps are now just in portal-proto.

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
aacic: Moved.

packages/portal-proto/src/components/Table/ExpandRowComponent.tsx
aacic: I've changed it to `value`, it works.

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
aacic: Defined `IDCParquetData` data type.

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
aacic: Changed to:  

`const idcTableColumnHelper = createColumnHelper<IDCViewerRow>();`

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
aacic: Fixed.

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
aacic: Changed to:

` const idcTableColumnHelper = createColumnHelper<IDCViewerRow>();`

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
aacic: Extracted to `IDCStudyRowComponent`

packages/portal-proto/src/components/Table/ExpandRowComponent.tsx
```diff
@@ -1,49 +1,53 @@
...
-            isRowExpanded && isColumnExpanded && "font-bold"
-          }`}
-        >
-          {value.length.toLocaleString().padStart(6)} {title}
-        </span>
-      </div>
-    )}
-  </>
-);
+  renderSingleValueInline?: boolean;
```
philtom-ctds: This flag argument has code smell.  Flag arguments are often indicators that something else should have been done. For example, if this were OOP, a flag argument that meant "perform behavior A if the flag is set" and "perform behavior B if the flag is not set" would be an indicator that the class should be split into two classes. Something similar should be done here.

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
```diff
@@ -0,0 +1,505 @@
...
+
+  // server-side pagination state
+  const [pageSize, setPageSize] = useState(20);
+  const [activePage, setActivePage] = useState(1);
+
+  // store API pagination metadata returned from triggerGetCases
+  const [apiPagination, setApiPagination] = useState<any>({});
+
+  // RTK Query lazy hook to fetch cases when we have the IDC case_ids available
+  const [triggerGetCases] = useLazyGetCasesQuery();
```
amlehman: These lazy queries are really just meant to be used in response to a user action like a logging in, requesting a download, etc. We should be using the standard hook when we are handling applications actions. I think the data fetching code needs some larger re-architecting to take advantage of RTK query hooks in the correct way. The parquet file is also requested/processed every time the cohort changes even though that's not necessary.

I think the request flow should look something like this:
```  
  const [idcData, setIdcData] = useState<{
    idc_data: readonly IDCParquetData[];
    case_ids: readonly string[];
  }>();


// On mount, fetch the parquet file
  useEffect(() => {
    const loadIdcData = async () => {
      const resp = await fetch(IDC_PARQUET_URL);
      if (!resp.ok) {
        throw new Error(
          `Failed to fetch parquet: ${resp.status} ${resp.statusText}`,
        );
      }

      const arrayBuffer = await resp.arrayBuffer();
      const idcData = await readParquetIndex(arrayBuffer);
      setIdcData(idcData);
    };

    loadIdcData();
  }, []);

  const extendedFilters = useDeepCompareMemo(() => {
    const id_filters: GqlOperation = {
      op: "in",
      content: {
        field: "case_id",
        value: idcData?.case_ids,
      },
    };

    const caseFiltersArg = cohortGqlFilters
      ? convertFilterToGqlFilter(cohortGqlFilters)
      : undefined;

    return caseFiltersArg
      ? mergeWithAndFilters(caseFiltersArg, id_filters)
      : id_filters;
  }, [idcData?.case_ids, cohortGqlFilters]);

  const sortByForApi = mapSortingToSortBy();

  const { data, isLoading, isError } = useGetCasesQuery(
    {
      request: {
        fields: ["submitter_id", "disease_type", "primary_site", "project"],
        size: pageSize,
        from: (activePage - 1) * pageSize,
        case_filters: extendedFilters,
        expand: ["samples.portions.slides", "project.program"],
        sortBy: sortByForApi.length > 0 ? sortByForApi : undefined,
      },
      fetchAll: false,
    },
    // this makes sure the cases request is fired sequentially after the parquet file request
    { skip: idcData === undefined },
  );

  useDeepCompareEffect(() => {
    const allHits = data?.hits || [];

    // capture API pagination metadata so we can render pagination controls
    setApiPagination(data?.pagination || {});

    // build mappings from allHits and idc_data
    const mappings = allHits.map((gdcCase: any) => {
      const submitterId = gdcCase?.submitter_id;
      const matches = (
        idcData.idc_data as ReadonlyArray<IDCParquetData>
      ).filter(
        (row: IDCParquetData) =>
          row.PatientID &&
          submitterId &&
          row.PatientID.toString() === submitterId.toString(),
      );
      return {
        gdcCase,
        matches,
      };
    });

    setMappings(mappings);
  }, [data, idcData]);

// update the rest of the app to use isLoading and isError and handle showing a loading state on mount

File: packages/portal-proto/src/components/Table/ExpandRowComponent.tsx [hunk 28]
```diff
@@ -1,49 +1,53 @@
-import React, { JSX } from "react";
+import React from "react";
 import { CollapseCircleIcon, ExpandCircleIcon } from "@/utils/icons";
 
 const ExpandRowComponent = ({
```
craigrbarnes, line 4: Recommend creating a custom component. This component is used 7 times with the default flag and once with the flag set to true, indicating it would be better as a custom component. Other apps in the portal, such as the Gene table in the MutationFrequency app, have a custom ExpandedRowComponent.

aacic: I've introduced `IDCExpandRowComponent` and reverted `ExpandRowComponent`.

packages/portal-proto/src/components/Table/ExpandRowComponent.tsx
aacic: I've introduced `IDCExpandRowComponent` and reverted `ExpandRowComponent`.

packages/portal-proto/src/components/Table/ExpandRowComponent.tsx
aacic: I've introduced `IDCExpandRowComponent` and reverted `ExpandRowComponent`.

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
aacic: I've refactored the code as suggested, and now the `useGetCasesQuery` hook is used.

packages/core/src/features/cases/casesSlice.ts
```diff
@@ -170,8 +170,8 @@ const slice = createSlice({
           case_id: datum.submitter_id,
           case_uuid: datum.case_id,
           submitter_id: datum.submitter_id,
-          project_id: datum.project.project_id,
-          program: datum.project.program.name,
+          project_id: datum.project?.project_id,
+          program: datum.project?.program?.name,
```
wteouchicago: did you run into issues with project and program here? same with summary below.

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
```diff
@@ -0,0 +1,496 @@
...
+    fetchAll: false,
+  });
+
+  // combined visibility for the LoadingOverlay: parquet load or remote fetch
+  const overlayVisible = parquetLoading || casesLoading;
+
+  // surface API errors to the shared loadError state so the UI can render
+  useEffect(() => {
+    if (casesError) {
+      setLoadError(
```
amlehman: It doesn't look like the value of 'loadError' is actually read (lines 437 - 444). Why not just use the a boolean flag?

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
amlehman: Looks great, thank you.

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
aacic: Good catch. I've changed it to a boolean.

packages/core/src/features/cases/casesSlice.ts
aacic: I indeed had some issues when I was using the `fetchGdcCases` directly, but now, when I use the hook  `useGetCasesQuery`, it works as expected with the reverted code, so I pushed it.

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
aacic: I've changed it to `useDeepCompareMemo` and it seems it works fine.

amlehman (approved):

packages/portal-proto/src/features/idc/IDCStudyRowsComponent.tsx
```diff
@@ -0,0 +1,93 @@
...
+                : null;
+            const radioLink =
+              s.hasRadiology && s.StudyInstanceUID
+                ? CT_VIEWER_BASE +
+                  "?StudyInstanceUIDs=" +
+                  encodeURIComponent(s.StudyInstanceUID)
+                : null;
+            return (
+              <tr
+                key={i}
```
paribartandhakal:
```suggestion
                key={s.StudyInstanceUID || i}
```

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
```diff
@@ -0,0 +1,481 @@
...
+  // Global loading state for parquet download/parsing + GDC fetch
+  const [parquetLoading, setParquetLoading] = useState<boolean>(false);
+  const [mappings, setMappings] = useState<any[]>([]);
+  const [loadError, setLoadError] = useState<boolean>(false);
+  const [expandedCases, setExpandedCases] = useState<Set<string>>(new Set());
+  // server-side pagination state
+  const [pageSize, setPageSize] = useState(20);
+  const [activePage, setActivePage] = useState(1);
+  // store API pagination metadata returned from triggerGetCases
+  const [apiPagination, setApiPagination] = useState<any>({});
```
paribartandhakal: I noticed we are storing the `apiPagination` data in `useState` and updating it via `useDeepCompareEffect`. Since this data comes directly from casesResponse, we can remove the state and effect entirely and just derive it directly like this:

```
const apiPagination = casesResponse?.pagination || {};
const pagination = {
  size: apiPagination?.size ?? pageSize,
  page: apiPagination?.page ?? activePage,
  // ...
};
``

File: packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx [hunk 33]
```diff
@@ -0,0 +1,481 @@
...
+    case_ids: gdcCaseIds,
+  };
+}
+
+const IDCViewerWrapper: FC = () => {
+  // Global loading state for parquet download/parsing + GDC fetch
+  const [parquetLoading, setParquetLoading] = useState<boolean>(false);
+  const [mappings, setMappings] = useState<any[]>([]);
+  const [loadError, setLoadError] = useState<boolean>(false);
+  const [expandedCases, setExpandedCases] = useState<Set<string>>(new Set());
```
paribartandhakal: I feel like this logic can be simplified. VerticalTable is made using Tanstack Table. 

TanStack table natively expects an object for its expanded state (`type ExpandedState = true | Record<string, boolean>`), we can simplify this quite a bit by removing the `Set` entirely and just storing the object directly. These changes will let us delete `toggleExpanded` function and the `expandedState` memo completely.

It would look something like:
```
// import ExpandedState from tanstack
const [expanded, setExpanded] = useState<ExpandedState>({});

const setTableExpanded = useCallback((row: Row<IDCViewerRow>) => {
    if (!row?.id) return;
    
    setExpanded((prev) => ({
      ...prev,
      [row.id]: !prev[row.id],
    }));
  }, []);

```
Then pass `expanded={expanded}` in VerticalTable below.

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
```diff
@@ -0,0 +1,481 @@
...
+
+  // When either the cases data or parquet idc data changes, rebuild mappings.
+  useDeepCompareEffect(() => {
+    const allHits = casesResponse?.hits || [];
+
+    // capture API pagination metadata so we can render pagination controls
+    setApiPagination(casesResponse?.pagination || {});
+
+    // build mappings from allHits and idc_data
+    const mappings = allHits.map((gdcCase: any) => {
```
paribartandhakal: just noticed this bit where are are iterating over the whole data on case data or parquet idc data change. would it be better to create a mapping of the idcData first and then look it up? That way we can also get rid of useEffect and useState.

This is just a suggestion:
```
const idcDataByPatientId = useMemo(() => {
    const map: Record<string, IDCParquetData[]> = {};
    if (!idcData?.idc_data) return map;

    idcData.idc_data.forEach((row) => {
      const pid = row.PatientID?.toString();
      if (!pid) return;
      if (!map[pid]) map[pid] = [];
      map[pid].push(row);
    });
    
    return map;
  }, [idcData]);

const mappings = useMemo(() => {
    const allHits = casesResponse?.hits || [];
    
    return allHits.map((gdcCase: any) => {
      const submitterId = gdcCase?.submitter_id?.toString();
      // instant lookup here
      const matches = submitterId ? (idcDataByPatientId[submitterId] || []) : [];
      
      return {
        gdcCase,
        matches,
      };
    });
  }, [casesResponse?.hits, idcDataByPatientId]);
```

I am not quite sure the size of idcData to say if the original one causes a delay or not.

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
```diff
@@ -0,0 +1,481 @@
...
+  "StudyInstanceUID",
+  "StudyDate",
+  "StudyDescription",
+  "study_type",
+  "gdc_case_id",
+];
+
+// Type for a single row read from the IDC parquet index.
+type IDCParquetData = {
+  PatientID: string;
```
paribartandhakal: Nitpick: we tend to use Camel Case for most of our variable names and types.

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
paribartandhakal: Also, I got tripped up here a bit between PatientId and submitter_id.

Should we add a comment about it that they are the same values?

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
aacic: Fixed.

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
aacic: Changed.

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
aacic: Good catch. I'm creating a mapping of the IDC data first, just the idc parquet file has just 17 000 rows, so it doesn't work much faster. But in case the table gets bigger, there will be a speed improvement. 

I've added a comment regarding PatientId and submitter_id as well.

paribartandhakal (commented):
I am okay with the changes code wise. I didn't verify the actual IDC viewer app functionality.

aacic:
> I am okay with the changes code wise. I didn't verify the actual IDC viewer app functionality.

Thank you! I was making sure that the code works after each change. Himanso will verify the functionality as well.

@paribartandhakal please approve the PR.

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
```diff
@@ -348,11 +348,7 @@ const IDCViewerWrapper: FC = () => {
                 }
               }}
               title={`${count} ${titleLabel}`}
-              role="button"
-              tabIndex={0}
-              aria-expanded={Boolean(info.row.getIsExpanded())}
-              aria-label={`${count} ${titleLabel}. Press Enter or Space to expand.`}
-              className="inline-block"
+              role="none"
```
paribartandhakal: What was the issue with axe here? These changes don't look right to me.

paribartandhakal: use `button` instead of `div`. Add back aria attributes. You can drop `onKeyDown` after using `button`. remove redundant `cursor-pointer` from `IDCExpandRowComponent` if not needed.

paribartandhakal: also, `info.row.getIsExpanded()` should already return a boolean.

siosonel (approved):

packages/portal-proto/src/features/idc/IDCViewerWrapper.tsx
aacic: I've reverted the changes. I don't see the axe issue anymore.  I was getting this error:

"Ensure interactive controls are not nested as they are not always announced by screen readers or can cause focus problems for assistive technologies"

Might be that there was a browser cache issue.

PR #1705, opened 2026-04-08
Fix for UUID issue in qa-int

paribartandhakal:
## Description
Hoist uuid to the top. currently, core and portal-proto has its own uuid. Since we are not copying core's node module in docker, it's failing. Now it's at the root level.

## Checklist
- [ ] Added proper unit tests
      flagged issues

### DIFF
--- a/packages/portal-proto/next-env.d.ts
+++ b/packages/portal-proto/next-env.d.ts
@@ -1,6 +1,0 @@
-/// <reference types="next" />
-/// <reference types="next/image-types/global" />
-import "./.next/types/routes.d.ts";
-
-// NOTE: This file should not be edited
-// see https://nextjs.org/docs/pages/api-reference/config/typescript for more information.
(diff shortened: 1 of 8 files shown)


packages/portal-proto/tsconfig.json
```diff
@@ -9,7 +9,7 @@
     "noEmit": true,
     "esModuleInterop": true,
     "module": "esnext",
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
```
paribartandhakal, line 12: auto updated by nextjs on build

packages/portal-proto/next-env.d.ts
```diff
@@ -1,6 +0,0 @@
-/// <reference types="next" />
-/// <reference types="next/image-types/global" />
```
paribartandhakal, line 2: I added global.d.ts and removed this as I read this shouldn't be commited. this is auto generated.

amlehman (approved):

packages/portal-proto/tsconfig.json
craigrbarnes: That seems reasonable since you changed the tsconfig.json

craigrbarnes (approved):

PR #1710, opened 2026-04-28
Improve idc search

aacic:
## Description

Fixes JIRA issues:

https://gdc-ctds.atlassian.net/browse/SV-2797

and

https://gdc-ctds.atlassian.net/browse/SV-2798

## Checklist

- [x] Added proper unit tests
- [x] Left proper TODO messages for any remaining tasks
- [x] Scanned for web accessibility with **aXe**, and mitigated or documented
      flagged issues

PR #1711, opened 2026-04-29
Parquet file versioning

aacic:
## Description

Resolves task:

https://gdc-ctds.atlassian.net/browse/FEAT-914

The checksum is still not added to the Parquet file, so that logic is not implemented yet.

The MD5 checksum is not added to the Parquet file yet, so the checksum check is not implemented. 

## Checklist

- [x] Added proper unit tests
      flagged issues

### DIFF
--- a/packages/portal-proto/src/features/proteinpaint/idc/utils.ts
+++ b/packages/portal-proto/src/features/proteinpaint/idc/utils.ts
@@ -1,0 +1,93 @@
+import { compressors } from "hyparquet-compressors";
+import { parquetReadObjects } from "hyparquet";
+import { IDCParquetData } from "@/features/proteinpaint/idc/types";
+
+export const IDC_BUCKET_URL =
+  "https://storage.googleapis.com/idc-index-data-artifacts/";
+export const IDC_PARQUET_KEY_SUFFIX =
+  "/release_artifacts/gdc_idc_mapping.parquet";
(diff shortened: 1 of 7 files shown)


aacic:
We need to merge this PR first to avoid conflicts:

https://github.com/NCI-GDC/gdc-frontend-framework/pull/1710

packages/portal-proto/src/features/proteinpaint/idc/utils.ts
```diff
@@ -0,0 +1,237 @@
...
+    const data = await loadParquetFromUrl(url);
+    return {
+      ...data,
+      metadataVersion: data.dataVersion,
+      urlVersion: version,
+      url,
+    };
+  } catch (err) {
+    console.warn(
+      `[IDC] Parquet artifact "${version}" is unavailable, inaccessible, or invalid and will be skipped:`,
```
amlehman, line 197: Is there any need to warn here given that the code is already handling the fallback? We generally try and avoid output to the console.

craigrbarnes: We do have a few warnings in the core, so we could let it in and then open a ticket to handle them later.

craigrbarnes (approved):
Looks fine to me, new packages do not introduce any security issues

PR #1718, opened 2026-05-19
PEAR-2639 - Update OICR's survival analysis plot on the FE

amlehman:
## Description
- Adds survival plot as a package in the gdc-frontend similar to the body plot
- Updated survival plot code to use recent versions of d3 & lodash instead of ancient ones :/
- Made viz changes

Before:
Screenshot 2026-05-19 at 11 46 32 AM

After:
Screenshot 2026-05-19 at 11 45 41 AM

## Checklist

- [ ] Added proper unit tests
      flagged issues


packages/survivalplot/src/index.js
```diff
@@ -0,0 +1,298 @@
...
+    top: 20,
+    right: 20,
+    bottom: 46,
+    left: 60,
+  },
+  shouldShowConfidenceIntervals: true,
+  confidenceAreaOpacity: 0.2,
+};
+
+const interpolation = d3.curveStepAfter;
```
amlehman, line 25: This is the change to how the points are connected (it was `d3.curveStepBefore`)

packages/survivalplot/src/index.js
```diff
@@ -0,0 +1,298 @@
...
+
+    // Draw the data as an svg path
+    setGroup
+      .append("svg:path")
+      .data([sampledDataPoints])
+      .attr("class", "line")
+      .attr("d", line)
+      .attr("stroke", setColor);
+
+    setGroup
```
amlehman: This (224-231) is the change to add the "zero line"

wteouchicago:
looks good! 👍 📉 👍

packages/survivalplot/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "@nci-gdc/survivalplot",
+  "version": "2.19.0",
+  "browser": true,
+  "main": "dist/index.js",
+  "module": "dist/index.esm.js",
+  "unpkg": "dist/index.umd.js",
+  "scripts": {
+    "clean": "rimraf dist",
+    "build": "rollup --config rollup.config.mjs"
```
paribartandhakal: I ran into issues running the app (couldn't find the package/dist). Should we be adding 
```
"build:watch": "npm run build -- --watch",
"dev": "npm run build:watch"
```

just like how `@nci-gdc/sapien` does it? So that `lerna` can pick up `dev` and build the package before `portal-proto` starts when we run `npm run dev` from the root.

packages/survivalplot/src/index.js
```diff
@@ -0,0 +1,298 @@
...
+      .append("line")
+      .attr("class", "line")
+      .attr("stroke", setColor)
+      .attr("x1", 0)
+      .attr("y1", y(sampledDataPoints[0].survivalEstimate))
```
paribartandhakal: We should probably make sure that the there is at least 1 `sampledDataPoints` before doing this, right?

packages/survivalplot/package.json
paribartandhakal: It worked for me after adding these.

packages/survivalplot/src/index.js
wteouchicago: oh, i should have specified in the ticket -- we definitely don't want the starting horizontal line if there are no case data points to plot. (that said, we currently also don't display the curve if there are under 10(?) data points -- not sure if that makes a diff).

amlehman: Yeah, I don't think this code will ever be hit in that scenario. But I can add a conditional just in case.

paribartandhakal (approved):

craigrbarnes (approved):

PR #1726, opened 2026-06-04
PEAR-2672: Fix for portal reloading on auth + Improve build and dependencies across the app (1st phase)

paribartandhakal:
## Description

The issue causing the Portal to hard reload during authentication was traced to our use of `local-ssl-proxy`, which we use to proxy the application over HTTPS. That package has not been updated in quite some time, so my suspicion is that the problem is related to changes in Node.js and/or stricter browser security policies. While investigating, I noticed that the WebSocket connection was being re-established approximately every 60 seconds, which appeared to be triggering the reload behavior.

As a solution, I replaced the `local-ssl-proxy` setup with the experimental HTTPS support now available through the Next.js CLI. This simplifies the local HTTPS setup, resolves the hard reload issue, and reduces our dependency on an unmaintained package.

Along the way, I also:

* Fixed several build dependency issues.
* Simplified the local development workflow.
* Added dedicated commands for running the Portal with and without HTTPS.
* Switched local auth testing to use locally generated certificates via `mkcert`.

For normal development, it is now much easier to run `npm run dev:gdc`, which serves the Portal over HTTP at `http://localhost:3000`.

When HTTPS is needed for authentication or login testing, you can run `npm run dev:gdc:https`, which serves the Portal at `https://localhost.gdc.cancer.gov:3000`.

### Local Auth Setup

To test login and other authentication-related functionality locally:

1. Configure your machine to map `localhost.gdc.cancer.gov` to `127.0.0.1`.

   Mac/Linux (`/etc/hosts`):

   ```text
   127.0.0.1 localhost localhost.gdc.cancer.gov
   ```

2. Install and configure `mkcert` (one-time setup):

   ```bash
   brew install mkcert
   mkcert -install
   ```

3. Generate local certificates from the repository root:

   ```bash
   mkdir -p .certs

   mkcert \
     -key-file .certs/localhost-key.pem \
     -cert-file .certs/localhost.pem \
     localhost.gdc.cancer.gov
   ```

4. Start the HTTPS development server:

   ```bash
   npm run dev:gdc:https
   ```

5. Open Chrome with web security disabled to suppress local CORS warnings:

   ```bash
   open -n -a /Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome \
     --args --user-data-dir="/tmp/chrome_dev_test" \
     --disable-web-security \
     https://localhost.gdc.cancer.gov:3000
   ```

6. Login and authentication flows should now work against:

   ```text
   https://localhost.gdc.cancer.gov:3000
   ```

### Known Behavior

After authenticating through the NIH login flow, the browser is redirected to:

```text
https://portal.gdc.cancer.gov/?login=true
```

The login tab does not automatically close. You must manually close that tab and return to the original local Portal tab.

Depending on timing, a manual refresh of the local Portal page may be required before the session cookie is detected and the UI reflects the authenticated state.

### Notes

* The `.certs/` directory is gitignored.
* Each developer must generate their own local certificates once using the `mkcert` steps above.
* Docker builds were verified successfully and the Portal loads correctly in local development.


wteouchicago:
Works great for me -- no more nasty reload every minute. 👍

amlehman (approved):

craigrbarnes (approved):

PR #1696, opened 2026-03-24
PEAR-2554: Tailwind v4 upgrade

paribartandhakal:
## Description
Upgrade to tailwind `4.2.2`. Still using tailwind config as we have lots of color already defined there. I read it's still performant.

## Checklist
- [ ] Added proper unit tests
- [ ] Left proper TODO messages for any remaining tasks
- [ ] Scanned for web accessibility with **aXe**, and mitigated or documented
      flagged issues

PR #1730, opened 2026-06-09
PEAR-2698 - cDAVE, Mutation Frequency - loading frame and hints overlap content when scrolling

amlehman:
## Description
I put back the baseZIndex prop that was removed in https://github.com/NCI-GDC/gdc-frontend-framework/pull/1696. The `Popover` component has a default z index of 300 so that was causing it to overlap with elements when it wasn't in a modal. I'm going to write a ticket to create a proper z index hierarchy but I wanted to get this fixed without touching the whole site first. 

## Checklist

- [ ] Added proper unit tests
      flagged issues

### DIFF
--- a/packages/portal-proto/src/components/InputEntityList/InputEntityList.tsx
+++ b/packages/portal-proto/src/components/InputEntityList/InputEntityList.tsx
@@ -314,6 +314,7 @@ const InputEntityList: React.FC<InputEntityListProps> = ({
               events={{ hover: true, focus: true, touch: false }}
               withArrow
               withinPortal={false}
+              zIndex={302}
             >
               <ActionIcon
                 data-testid="tooltip-accepted-identifier-info"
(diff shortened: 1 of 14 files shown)


paribartandhakal (approved):
Looks good. Also, setting up the hierarchy sounds great. That would standardize the implementation and make it less error prone. Thank you for fixing this.

wteouchicago:
Oddly in the Mutations table in Mutation Frequency, the search box's tooltip still floats above the cohort bar. Not a big deal if we don't fix though.

Screenshot 2026-06-10 at 1 22 01 PM

Same with the following, but they are also happening in prod, so not a regression and not necc to fix.
- Project summary page, Annotations table's search tooltip and Primary Site's search tooltip
- Case summary page, Files and Annotations (this case has lots of annotations: 7248cd60-be22-44bc-bc58-f644db0940a2)

All of these can wait until the upcoming One True Fix[tm].

PR #1750, opened 2026-07-22
PEAR-2576 - Investigate and possibly implement a better way to handle removed properties

amlehman:
## Description
I stole your idea Wendy, works great.

## Checklist

- [ ] Added proper unit tests
      flagged issues


packages/core/src/listeners.ts
```diff
@@ -167,3 +172,22 @@ startCoreListening({
...
 });
+
+/* Remove fields that aren't in the facet dictionary from user's saved custom facets */
+startCoreListening({
+  matcher: isAnyOf(fetchFacetDictionary.fulfilled),
+  effect: async (action, listenerApi) => {
+    if (isPayloadActionWithObject(action)) {
+      const validFields = Object.keys(action.payload);
+      const customFacets = selectBuilderCustomFacets(listenerApi.getState());
+      const invalidFacet = customFacets.filter(
```
paribartandhakal: I think we should rename this to `invalidFacets` for clarity and use the same name below.

```suggestion
      const invalidFacets = customFacets.filter(
```

packages/core/src/listeners.ts
```diff
@@ -167,3 +172,22 @@ startCoreListening({
...
+        (customFacet) => !validFields.includes(customFacet),
+      );
+      invalidFacet.forEach((invalidFacet) =>
```
paribartandhakal:
```suggestion
      invalidFacets.forEach((invalidFacet) =>
```

wteouchicago:
@amlehman Some questions (I haven't tested yet):
- would this change work for the Repository's custom filters as well? 
- if ^^ not, is it possible to make that work? Otherwise, we would still need to maintain the hardcoded list.
- if ^^ so, would we need to remove the `deprecatedFields.ts` file and related stuff that depends on it? 🤷‍♀️

packages/core/src/features/gdcapps/createAppStore.ts
```diff
@@ -0,0 +1,68 @@
+import { createContext } from "react";
```
amlehman, line 1: I noticed when I was working on the code in [‎packages/portal-proto/src/features/repositoryApp/subscribers.ts‎](https://github.com/NCI-GDC/gdc-frontend-framework/pull/1750/changes#diff-2085046e811829a4fefbbe7363e100151ed39ce7f3b3cebbd6014632df106492) that the type hinting was not working for the gdcapp hooks. This change is to fix that.

amlehman:
Hi @wteouchicago, I was able to get something similar working for the repository filters (it was a little harder since they are stored in different places) and to remove the old migrations.

packages/core/src/listeners.ts
amlehman: Good catch, I've updated this name.

wteouchicago:
Works great for me! 👍 I tested by doing the following:
- added custom filters in Cohort Builder and Repository: 2 in each
- in local storage and session storage, modify the name of one property for each app to a nonexistent prop 
- refresh
- verify that only one card is left in the custom filters and it corresponds to the property I didn't edit

paribartandhakal (approved):

packages/portal-proto/src/features/annotations/appApi.ts
```diff
@@ -11,7 +11,7 @@ const reducers = combineReducers({
 });
 
 export const { id, AppStore, AppContext, useAppSelector, useAppDispatch } =
-  createAppStore({
+  createAppStore<AppState>({
```
craigrbarnes, line 14: Nice

craigrbarnes (approved):
Changes look fine

PR #1759, opened 2026-08-12
PEAR-2709 - Investigate possibility of reducing calls to /cohorts

amlehman:
## Checklist

- [ ] Added proper unit tests
      flagged issues

### DIFF
--- a/packages/portal-proto/src/features/cohortBuilder/hooks.ts
+++ b/packages/portal-proto/src/features/cohortBuilder/hooks.ts
@@ -2,6 +2,7 @@ import { useState, useEffect } from "react";
 import { useDeepCompareEffect, useDeepCompareMemo } from "use-deep-compare";
 import { flatten } from "lodash";
 import MiniSearch from "minisearch";
+import Cookies from "js-cookie";
 import {
   useCoreDispatch,
   useCoreSelector,
@@ -38,11 +39,14 @@ import { FacetQueryOptions } from "../facets/types";
 
 export const useSetupInitialCohorts = (): boolean => {
   const [fetched, setFetched] = useState(false);
+  // If the user doesn't have a context id, then we should skip the call to cohorts because it will return an error
+  const contextId = Cookies.get("gdc_context_id");
+  const skipped = contextId === undefined;
   const {
     data: cohortsListData,
     isSuccess,
     isError,
-  } = useGetCohortsByContextIdQuery(null, { skip: fetched });
+  } = useGetCohortsByContextIdQuery(null, { skip: fetched || skipped });
 
   const coreDispatch = useCoreDispatch();
   const cohorts: Cohort[] = useCoreSelector((state) => selectAllCohorts(state));
@@ -110,7 +114,7 @@ export const useSetupInitialCohorts = (): boolean => {
     cohortWarnings,
   ]);
 
-  return fetched;
+  return fetched || skipped;
 };
 
 export const usePopulateFacetData = (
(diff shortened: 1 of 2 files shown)


packages/portal-proto/src/features/cohortBuilder/hooks.ts
```diff
@@ -35,14 +36,18 @@ import { useEnumFacets } from "@/features/facets/hooks";
 import { STOP_WORDS, TOKENIZE_STRING } from "./dictionary";
 import { useCohortFacetFilters } from "./utils";
 import { FacetQueryOptions } from "../facets/types";
+import { isDevEnvironment } from "src/pages/_app";
 
 export const useSetupInitialCohorts = (): boolean => {
   const [fetched, setFetched] = useState(false);
+  // If the user doesn't have a context id, then we should skip the call to cohorts because it will return an error
+  const contextId = Cookies.get("gdc_context_id");
+  const skipped = contextId === undefined && !isDevEnvironment();
```
amlehman, line 45: We can't read this cookie on `localhost` development sites (i.e it comes back undefined everytime) because the domain doesn't match. So I added this check to make sure we aren't always skipping this request on development sites. This feels a little messy to me though :(

amlehman: Oh this also means it isn't testable on `localhost` development sites, it will have to be tested on my dev cluster.

wteouchicago: > We can't read this cookie on localhost development sites (i.e it comes back undefined everytime) because the domain doesn't match. So I added this check to make sure we aren't always skipping this request on development sites. This feels a little messy to me though :(

Oh, is that why I didn't have the context stored in local storage when I was on localhost?

wteouchicago:
Looks ok testing locally just to make sure cohorts are still being pulled when there is a context id.

@amlehman -- whenever you get the chance, maybe you could put this on your dev cluster and I can continue testing there?

packages/portal-proto/src/features/cohortBuilder/hooks.ts
amlehman: Yes, the code that stores the context id in local storage doesn't work on localhost sites as well.

craigrbarnes: While it's nice to have the same code for development and production, I think this is an acceptable case.

However, can you add a comment: _This isn't testable on localhost development sites; it will have to be tested on a dev cluster._

amlehman:
Hi @wteouchicago, this should be deployed to my dev cluster.

wteouchicago:
Works great! Not seeing that unnecessary /cohorts call anymore when I don't have a context id. 👍

packages/portal-proto/src/pages/_app.tsx
```diff
@@ -47,6 +47,8 @@ if (typeof window !== "undefined" && process.env.NODE_ENV !== "production") {
   axe(React, ReactDOM, 1000);
 }
 
+export const isDevEnvironment = () => process.env.NODE_ENV !== "production";
```
paribartandhakal, line 50: Just wondering why does this need to be defined in this page?

amlehman: This is just where we are doing other reading of environment variables. Open to suggestions for a better place for it.

paribartandhakal: Hmm okay. Yeah, ideally we could make a new file too keep these. Just noticed that `_app.tsx` needs a bit more work and clean up.

paribartandhakal (approved):

craigrbarnes (approved):

PR #1769, opened 2026-09-04
rewrite SJ tool wrapper imports of GFF code to minimize impacts to test code

siosonel:
## Description

This branch addresses https://gdc-ctds.atlassian.net/browse/FEAT-925. 

The main change is to select the specific @gff/core properties/functions that are used by SJ tool wrappers, and to export these limited properties from a `coreAdapter` that is simpler to mock and reuse across `proteinpaint/*.test.*` files. 

NOTES:
- The boundary mocks are inline factories, not automock. jest.mock("../cohortBuilder/utils") bare would automock — which loads the real module and re-triggers @gff/core. The factories (`() => ({ INVALID_COHORT_NAMES: [] })`) are what actually keep @gff/core out. These can become shared __mocks__ files, but factories keep each test self-documenting about why the boundary exists.
- What still legitimately breaks the tests: if GFF renames a hook a wrapper directly uses, the wrapper stops compiling and coreAdapter.ts needs one edit — that's real integration surface, now a single obvious file rather than nine. 

## Checklist

- [ ] Added proper unit tests
      flagged issues

### DIFF
--- a/packages/portal-proto/src/features/proteinpaint/CnvSegmentWrapper.unit.test.tsx
+++ b/packages/portal-proto/src/features/proteinpaint/CnvSegmentWrapper.unit.test.tsx
@@ -1,43 +1,30 @@
 import { render } from "@testing-library/react";
-import { ProteinPaintWrapper } from "./ProteinPaintWrapper";
 import { MantineProvider } from "@mantine/core";
+import * as coreAdapter from "./coreAdapter";
+import { ProteinPaintWrapper } from "./ProteinPaintWrapper";
 
 const filter = { abc: "xyz" };
 let runpparg,
   userDetails,
   isDemoMode = false;
 
-const resultsCreateCaseSet = { data: "test-pp-caseSet", isSuccess: true };
-const nullFunction = () => null;
+// (1) The single @gff/core seam for the wrapper's OWN core imports.
+jest.mock("./coreAdapter");
 
-jest.mock("@gff/core", () => ({
-  useCoreSelector: jest.fn().mockReturnValue({}),
-  buildCohortGqlOperator: jest.fn(() => filter),
-  useAddCohortMutation: jest.fn(() => [() => null, { isSuccess: true }]),
-  useFetchUserDetailsQuery: jest.fn(() => userDetails),
-  useCoreDispatch: jest.fn(() => nullFunction()),
-  setActiveCohort: jest.fn(() => null),
-  PROTEINPAINT_API: "host:port/basepath",
-  useCreateCaseSetFromValuesMutation: () => [
-    nullFunction,
-    resultsCreateCaseSet,
-  ],
-  useLazyGetCohortsByContextIdQuery: jest.fn().mockReturnValue([
-    jest.fn().mockReturnValue({ unwrap: jest.fn() }),
-    {
-      isSuccess: true,
-      isLoading: false,
-    },
-  ] as any),
-  useLazyGetCohortByIdQuery: jest.fn().mockReturnValue([jest.fn()]),
-  useCreateCaseSetFromFiltersMutation: jest.fn().mockReturnValue([jest.fn()]),
+// (2) Out-of-dir code the wrapper renders that reaches @gff/core on its own.
+jest.mock("@gff/portal-components", () => ({
+  SaveCohortModal: () => null,
+}));
+jest.mock("../cohortBuilder/CohortManager/cohortActionHooks", () => ({
+  cohortActionsHooks: {},
+}));
+jest.mock("../cohortBuilder/utils", () => ({
+  INVALID_COHORT_NAMES: [],
 }));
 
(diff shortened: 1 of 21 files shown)


amlehman:
I think the ask of FEAT-925 has been misunderstood here. We want to follow some best practices when writing these tests:

1) Test user behavior of a component not implementation details
2) Test the actual code as much as possible and avoid over mocking. Try to only mock out API interaction and the "existing state" of the store. 

I don't think this pattern of importing a huge amount of mocks for each test is a good idea. It runs in opposition to best practice <span>#</span>2. That file will only grow and I already see some mocks that shouldn't be necessary (`showModal` and `hideModal` for example). 

What we want to avoid is a cascade of mocks - where we are having to continually add more things to the list of mocked functions because we've mocked entire modules (like gff/core) out. That leads to continually maintaining these tests if we change the portal code that is imported by the proteinpaint apps. What I was hoping to see here was slimming down the amount of mocks to just those that are actually necessary. Here is an example of what I was thinking of for these changes: https://github.com/NCI-GDC/gdc-frontend-framework/compare/develop...less-mocked-tests

siosonel:
@amlehman Thanks for the review and example. The wrapper tests now use `...jest.requireActual("@gff/core")` in the mock per your example. Please let me know if your think other changes/improvements are necessary.

siosonel:
@craigrbarnes @philtom-ctds @wteouchicago not sure what the next steps here are, if there are still things that I need to work on

wteouchicago:
> @craigrbarnes @philtom-ctds @wteouchicago not sure what the next steps here are, if there are still things that I need to work on

Hi @siosonel -- our team will discuss and get back to you soon.

amlehman (approved):

craigrbarnes (approved):

packages/portal-proto/src/features/proteinpaint/CnvSegmentWrapper.unit.test.tsx
```diff
@@ -1,91 +1,46 @@
...
+import { useIsDemoApp } from "@/hooks/useIsDemoApp";
+import { getExpectedFilter0 } from "./ppTestHelpers";
 import { ProteinPaintWrapper } from "./ProteinPaintWrapper";
-import { MantineProvider } from "@mantine/core";
 
-const filter = { abc: "xyz" };
-let runpparg,
-  userDetails,
-  isDemoMode = false;
+let runpparg;
```
paribartandhakal: `runpparg` is defined globally but never reset. This means the second test could accidentally pass or fail based on data left over from the first test.

So it's a best habit to clear mocks and separate the concern. We could create a new ticket to tackle this.

This is happening in all the tests.

cc. @siosonel

packages/portal-proto/src/features/proteinpaint/CnvSegmentWrapper.unit.test.tsx
```diff
@@ -1,91 +1,46 @@
...
   expect(typeof runpparg.host).toBe("string");
   expect(runpparg.noheader).toEqual(true);
   expect(runpparg.nobox).toEqual(true);
   expect(runpparg.holder instanceof HTMLElement).toBe(true);
-  expect(runpparg.filter0).toEqual(filter);
+  expect(runpparg.filter0).toEqual(expectedFilter0);
   expect(runpparg.allow2selectSamples).toEqual({
     buttonText: "Create Cohort",
     attributes: [{ from: "sample_id", to: "cases.case_id", convert: true }],
     callback: runpparg.allow2selectSamples?.callback,
```
paribartandhakal: `callback: runpparg.allow2selectSamples?.callback` compares the value to itself, meaning it will always pass even if the callback is accidentally missing. It's safer to use callback: `expect.any(Function)`.

packages/portal-proto/src/features/proteinpaint/CnvSegmentWrapper.unit.test.tsx
```diff
@@ -1,91 +1,46 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  require("./ppTestHelpers").mockGffCore(),
+);
 
-jest.mock("@gff/portal-components");
+jest.mock("@/hooks/useIsDemoApp");
 
 jest.mock("@sjcrh/proteinpaint-client", () => ({
   __esModule: true,
   bindProteinPaint: jest.fn(async (arg) => {
```
paribartandhakal: Also, since `bindProteinPaint` fires asynchronously inside `useDeepCompareEffect`, checking `runpparg` synchronously can cause race conditions. It's usually safer to clear mocks between tests, use `waitFor`, and check `jest.mocked(bindProteinPaint).mock.calls` directly instead of using a global variable.

craigrbarnes (changes requested):
@siosonel Based on Pari's comment could you update these test?

